PR #26, opened 2015-06-25
using walker's override function

jordansread:


jordansread:
@ldecicco-USGS merge?

Issue #42, opened 2015-06-25
fix issue with calls to lines and points where you need to specify x= and y=


jordansread:
can't do:

``` R
points(gs, 1:5, 1:5) 
```

want to

jiwalker-usgs:
#50 should fix this, not the prettiest and has some manual copying that might cause headaches later

jordansread:
probably get a test in there that calls the examples in `graphics::points,lines`? worth it?

jiwalker-usgs:
yeah, examples should probably do gsplot and base graphics calls to illustrate how awesome everything is.

jordansread:
I will create a test and then close this issue

jordansread:
``` R
plot(-4:4, -4:4, type = "n")  # setting up coord. system
  points(rnorm(200), rnorm(200), col = "red")
```

``` barf
 Error in plot.xy(xy.coords(x, y), type = type, ...) : invalid plot type 
```

ldecicco-USGS:
So what is this test trying to show? Once you've put type="n"....I'd want an error when calling `points`

Plus, that's all `plot.xy`, right?

ldecicco-USGS:
I'd like tests, examples, and a kick-ass vignette too by the way

PR #51, opened 2015-06-26
broken test for #42

jordansread:
need a fix for #42


jordansread:
this is expected to fail BTW

Issue #45, opened 2015-06-25
plot margins should expand to accomodate legend


jiwalker-usgs:

thongsav-usgs:
Should be finished with #56

PR #55, opened 2015-06-29
Plots acomodate out-of-plot legends now

thongsav-usgs:


thongsav-usgs:
For issue https://github.com/USGS-R/gsplot/issues/45

thongsav-usgs:
Might conflict with @ldecicco-USGS's last PR, I can handle the merge if so. https://github.com/USGS-R/gsplot/pull/52

PR #60, opened 2015-07-01
side-specific lims

jordansread:
`axis` examples fail, but I am not sure what we want to do there...default lims? what happens with an empty plot?


ldecicco-USGS:
Yeah, I agree...I was suggesting earlier to take out that example.

ldecicco-USGS:
:shipit:

PR #64, opened 2015-07-02
fix for broken points() test

jordansread:
fix for points

we had `list(object)` before, but that was breaking the `axis` and `legend` calls (or giving warnings)


ldecicco-USGS:
Whoops...that was me.

PR #70, opened 2015-07-02
Abline

ldecicco-USGS:


ldecicco-USGS:
OK, I added an example to legends that breaks...that we need to fix. But otherwise, this starts to clean up the readme (using readme.Rmd), allows the gsplot(), passes the as-cran checks...and has the abline stuff in it.

ldecicco-USGS:
I fixed the x,y legend this morning so all checks are passing. It's a little clunky, probably worth refactoring the legend code with the same flow as points/lines etc (basically, make the argument list mirror the graphics::legend stuff). Anyway, I'm done with this for awhile...

jiwalker-usgs:
Lots of good stuff in here. :+1:

PR #84, opened 2015-07-08
Axis + error bars

ldecicco-USGS:
This pull request should probably have an actual look-over before merging.


R/calc_views.R
```diff
@@ -80,16 +80,47 @@ calc_view_usr <- function(views){
...
-      client_lims <- lims_from_client(side_components, var='ylim', side)
-      lims[!is.na(client_lims)] <- client_lims[!is.na(client_lims)]
-      usr <- usr_from_lim(lims, type=par()$yaxs)
+      usr <- usr_from_lim(lims, type=par()$yaxs) 
     } else {
       # is x
-      lims <- lims_from_list(lapply(side_components, var='x', function(list, var) strip_pts(list,var)))
-      client_lims <- lims_from_client(side_components, var='xlim', side)
-      lims[!is.na(client_lims)] <- client_lims[!is.na(client_lims)]
+      if(length(grep("^x$", unlist(lapply(side_components, names)))) > 0){
```
jordansread, line 105: you fancy

ldecicco-USGS, line 105: I think you mean hacky

R/overrideGraphics.R
```diff
@@ -14,11 +14,16 @@ overrideGraphics <- function(name, object, ...) {
         params <- append(list(object), params)
     }
     
-    
-    defFun <- getFromNamespace(ifelse(existsFunction(paste0(name,".default")), paste0(name,".default"), name), 'graphics')
+    base.package <- ifelse(length(grep(pattern="error_bar", x=name)) > 0, "gsplot", "graphics")
+    defFun <- getFromNamespace(ifelse(existsFunction(paste0(name,".default")), paste0(name,".default"), name), base.package)
```
jordansread, line 18: we should make sure these are the same things that happen in config

R/overrideGraphics.R
```diff
@@ -14,11 +14,16 @@ overrideGraphics <- function(name, object, ...) {
...
     
     names(params)[which(names(params) == "")] <- names(formals(defFun))[which(names(params) == "")]
     
-    do.call(getFromNamespace(name, "graphics"), params)
+    if(base.package == "gsplot"){
```
jordansread, line 22: why the change?

ldecicco-USGS, line 22: There was a recursive problem. It was pretty funky actually. So this is allowing us to make new functions (like error_bars), set a default and gsplot method.

R/overrideGraphics.R
ldecicco-USGS, line 18: hmm...yeah, in my head I made this more generalized...apparently not in real life

Issue #87, opened 2015-07-08
Take into account arrows for xylim calculations


mhines-usgs:

jordansread:
``` R
gs<- gsplot() %>% arrows(1,1,2,3) %>% points(1,1)
gs
```

jordansread:
does this:
![image](https://cloud.githubusercontent.com/assets/2349007/8581715/214a44fc-2588-11e5-8409-15648e98db61.png)
 we want it to do this:
![image](https://cloud.githubusercontent.com/assets/2349007/8581749/4cf29a46-2588-11e5-8249-46dd261815b0.png)

ldecicco-USGS:
I think change:
https://github.com/USGS-R/gsplot/blob/master/R/calc_views.R#L114

to something like:

``` r
xData <- do.call(c, unname(dataViews[ which(names(dataViews) %in% c("x","x1","x2")]))
```

and the same for y here:
https://github.com/USGS-R/gsplot/blob/master/R/calc_views.R#L92

ldecicco-USGS:
(UNTESTED!)

PR #90, opened 2015-07-08
fixes for #87

jordansread:
image


R/calc_views.R
```diff
@@ -80,8 +80,8 @@ calc_view_usr <- function(views){
     
     if ((side %% 2) == 0){
       # is y 
-      if(length(grep("^y$", unlist(lapply(side_components, names)))) > 0){
-        lims <- lims_from_list(lapply(side_components, var='y', function(list, var) strip_pts(list,var)))
+      if (any(unname(unlist(lapply(side_components, names))) %in% c('y','y1','y0'))){
+        lims <- lims_from_list(lapply(side_components, var=c('y','y1','y0'), function(list, var) strip_pts(list,var)))
```
ldecicco-USGS, line 84: I think you need to make the changes on line 92 too

R/calc_views.R
```diff
@@ -102,8 +102,8 @@ calc_view_usr <- function(views){
       usr <- usr_from_lim(lims, type=par()$yaxs) 
     } else {
       # is x
-      if(length(grep("^x$", unlist(lapply(side_components, names)))) > 0){
-        lims <- lims_from_list(lapply(side_components, var='x', function(list, var) strip_pts(list,var)))
+      if(any(unname(unlist(lapply(side_components, names))) %in% c('x','x1','x0'))){
```
ldecicco-USGS, line 105: And line 114

R/calc_views.R
mhines-usgs, line 84: @ldecicco-USGS do you have a minute to talk about these other places, i'm not totally sure I (we, maybe @lindsaycarr wants to join us too) understand what is going on with the code enough to edit it. :/

lindsayplatt, line 84: @mhines-usgs I would like to join you

ldecicco-USGS, line 84: Sure....but take a few minutes to hack around first. The problem is that in the 'if' statement we changed above, we realized we need to look for not just y, but y, y1, and y0 in 'side_components'. When side_components doesn't have y info (the 'else' statement), then we still need to calculate some limits (this is the condition that allows us to specify axes). Anyway, line 92 is looking for just y again....but we need it to look for y, y1, and y0. (and....same with x's on line 114). 

An example that needs to work is:

``` r
gs <- gsplot() %>%
  arrows(x0=1, y0=2, x1=2, y1=4, lwd=2) %>%
  axis(side=c(3,4),labels=FALSE) 
gs
```

Come find me when you've poked around a bit.

Issue #94, opened 2015-07-09
Background color


ldecicco-USGS:
I've always wanted a easy way to background color option to do something like this:

![2015-07-09_17-14-41](https://cloud.githubusercontent.com/assets/1105215/8608176/212c5902-265e-11e5-8953-1b91c5e24d0d.png)

ldecicco-USGS:
Workflow might be something like:

``` r
gs <- gsplot()%>%
  points(1,1) %>%
  bg("red")
gs
```

Base R is:

``` r
plot(1,1)
rect(par("usr")[1],par("usr")[3],par("usr")[2],par("usr")[4],col = "red")
par(new=TRUE)
plot(1,1)
```

lindsayplatt:
There's a function within ggplot2 that is called "panel.background" - it might be what you are looking for. This is the example I came up with:

``` r
df <- as.data.frame(cbind(c(1,1), c(2,2)))
p <- ggplot(df) + 
  geom_point(aes(x=df[,1], y=df[,2])) +
  theme(panel.background = element_rect(fill="red"))
p
```

ldecicco-USGS:
Yup, that's exactly the idea. We're not using ggplot2 however, so we need to make our own function.

ldecicco-USGS:
FYI, we're going supporting geom_point either (at least for now...I think that would be entirely too confusing for both users and developers to support both base R and ggplot2 functions)

lindsayplatt:
Right, ok not really sure where to start to create the function that fits in to the gsplot package...I know how to set up basic functions, but not sure about functions within a package

ldecicco-USGS:
It's not trivial. Have a look at the error_bars functions:

https://github.com/USGS-R/gsplot/blob/master/R/error_bars.R

I'm a little busy right this sec, but see if you can figure out what's going on there. It would be the same idea to add a function bgCol or something like that.

lindsayplatt:
Yeah, no worries. I'll see what I can crankout

ldecicco-USGS:
:+1:

lindsayplatt:
So I am not sure exactly how to set up this function to work with the rest of gsplot because I don't really understand how all of the functions in the package work together yet. I used something from a package called "RColorBrewer" for now to generate a list of colors in a gradient, but can remove that if the idea is to minimize the number of other packages used. 

Here is a function using RColorBrewer and base plot features: 

``` r
[...]

PR #98, opened 2015-07-11
bgCol

ldecicco-USGS:
Here's what I was thinking for #94


R/calc_views.R
```diff
@@ -19,6 +19,46 @@ calc_views <- function(gsplot){
   
   views <- add_view_labs(views, labs)
   
+  #Start enforcing some order. Background color has to go in back...I'd also put grid in back:
```
jordansread, line 22: I feel like we need a strategy for calc_views and some info in the repo's wiki about how to add new features to the package. 
My struggle is that we need to define the new functions in a bunch of different places - config, calc_views (sometimes), and in some cases, overrideGraphics()

jordansread:
:+1:

Issue #89, opened 2015-07-08
Create callout function?


mhines-usgs:
![calloutexample](https://cloud.githubusercontent.com/assets/6920344/8580692/b0578512-2581-11e5-9698-683b456c4e9d.jpg)

jordansread:
I'm on it.

PR #102, opened 2015-07-11
initial commit for #89

jordansread:
``` R
gs <- gsplot()
gsNew <- points(gs, y=1, x=2, xlim=c(0,3),ylim=c(0,3),
             col="blue", pch=18, legend.name="Points")
gsNew <- callouts(gsNew, 2, 1, labels='dog')
gsNew
```

image

There are a few to-do notes in the code comments


jordansread:
@ldecicco-USGS or @mhines-usgs take a peek?

jordansread:
looks like installs failed on the builds. GH issue?

mhines-usgs:
hah @jread-usgs -> :dog:

mhines-usgs:
This is really cool, Jordan! I'm going to merge so I can test this with your repgen changes locally.

jiwalker-usgs:
One thought I have here is similar to what we were talking with error bars.  Since these callouts may be tied to points, allow them to be declared within the points call?

jordansread:
yep, I think that makes sense moving forward. Still need the low level calls though too.

Issue #71, opened 2015-07-02
Log


ldecicco-USGS:
Support

jordansread:
highest priority :100:

jordansread:
Am working on this on a logged_axis branch. Challenging because we need to also reset the `xaxp` or `yaxp` if y or x is logged.

jordansread:
``` R
gs <- gsplot() %>%
    lines(c(1,3,4,20),c(1,2,3,4),log='x')
gs
```

![image](https://cloud.githubusercontent.com/assets/2349007/8641645/6b0157da-28de-11e5-965d-1e8244aba89e.png)

but I am hardcoding the `xaxp` reset manually right now until I can figure it out.

Issue #101, opened 2015-07-11
segments(x0=1, y0=2, 3, 4) fails


jordansread:
Error in (function (x0, y0, x1 = x0, y1 = y0, col = par("fg"), lty = par("lty"),  : 
  formal argument "x0" matched by multiple actual arguments

This ends up doing this to the args:

params
$x0
[1] 3

$x0
[1] 1

$y0
[1] 2

$y1
[1] 4

ldecicco-USGS:
Oh, I just fixed the same thing for the error bar functions. I'll get to this at some point (unless someone whats to look at what I did).

PR #105, opened 2015-07-14
code for #71 and tests too!

jordansread:


jordansread:
whoops, I meant "code for #101 and tests too!"

PR #110, opened 2015-07-15
removing dev.cur()==1 test

jordansread:
@lindsaycarr can you try this out?


lindsayplatt:
Works with the split.screen example from #94

jordansread:
@ldecicco-USGS you might want to look back on this when you have time. I pulled out the dev.cur(), but it would be good to discuss alternatives.

Issue #111, opened 2015-07-15
Line type and color changes only in legend


lindsayplatt:
When line types and colors are changed, it changes the appearance in the legend, but not in the actual plot:

``` r
gs <- gsplot() %>%
  points(x=2, y=2, col="blue", pch=22, legend.name="Points") %>%
  abline(h=2, col="red", lty=3, legend.name="Horizontal line") %>%
  legend("topleft", title="Legend")
gs
```

jordansread:
I think I broke something.

jiwalker-usgs:
:disappointed:

PR #112, opened 2015-07-15
fixes and test for #111

jordansread:
this is why seeing diffs in the readme files is something you shouldn't ignore (sorry all...) :-1:


jordansread:
@lindsaycarr can you check that this addresses the issue for you and merge if it does (and if you agree with the changes)?

PR #114, opened 2015-07-15
simplifying naming of windows

jordansread:
`windows` are what the view renders, vs `gs.config` are args that come from the function to set window properties. Separating these for simplicity and readability.


jordansread:
merging this since it is really just naming changes

PR #118, opened 2015-07-16
using override instead of goverride

jordansread:
@jiwalker-usgs


jiwalker-usgs:
has travis been failing lately?

R/overrideGraphics.R
```diff
@@ -1,29 +1,22 @@
-overrideGraphics <- function(name, object, ...) {
+override <- function(package, name, object, ...) {
```
jiwalker-usgs, line 1: Will probably want to rename this file since the function name changed.

jiwalker-usgs:
:shipit:

R/overrideGraphics.R
jordansread, line 1: yes, I will do that.

jordansread:
that travis issue looks like a sudo problem, which I thought I fixed...

NAMESPACE
```diff
@@ -7,7 +7,6 @@ export(abline)
 export(arrows)
 export(axis)
 export(bgCol)
-export(bgCol.default)
```
ldecicco-USGS, line 10: Why did this get deleted? And, if that works, can we lose our other custom function's .default exports? (error_bars, callouts)

jordansread, line 10: I deleted it because it seemed like a private function. Should we export those? I vote no but am up for doing what the group thinks is best.

ldecicco-USGS, line 10: I think I'm convinced it doesn't break anything,so no reason not to take out the other ones too, right?

ldecicco-USGS, line 10: OH...but then you don't know what the arguments are in the help files....I get it now. So....bgCol is unique because the only required variable is col. I think I'd actually vote to put it back in so we can include those arguments in the help files.

jordansread, line 10: good point. Let's discuss

Issue #14, opened 2015-06-24
inset plots


ldecicco-USGS:

ldecicco-USGS:
PR #95 now allows this:

``` r
gs <- gsplot() %>%
   points( y=c(3,1,2), x=1:3, xlim=c(0,NA),ylim=c(0,NA),
            col="blue", pch=18, legend.name="Points", xlab="Index") %>%
   lines(c(3,4,3), c(2,4,6), legend.name="Lines", ylab="Data") %>%
   abline(b=1, a=0, legend.name="1:1") %>%
   legend("topleft",title="Awesome!") %>%
   grid() %>%
   error_bar_vertical(x=1:3, y=c(3,1,2), y.high=c(0.5,0.25,1), y.low=0.1) %>%
   arrows(x0=0.75, y0=2, x1=1, y1=2.8, lwd=2) %>%
   title("Graphing Fun") %>%
   text(.75,1.75,labels="Weird data") %>%
   axis(side=c(3,4), labels=FALSE)
gs
u <- par("plt")
v <- c( (u[1]+u[2])/2, u[2], (u[3]+u[4])/2, u[4] )
options("gsplot"=config("par",new=TRUE,fig=v))
gs2 <- gsplot() %>%
   points(1, 2) %>%
  grid()
gs2
#Reset:
options("gsplot"=NULL)

```

ldecicco-USGS:
Which produces
![test](https://cloud.githubusercontent.com/assets/1105215/8608677/d093c41c-2662-11e5-8ae8-767c67155b31.png)
:

ldecicco-USGS:
So, not perfect, but a start

ldecicco-USGS:
Also the reason I'd want a background color function. Then the inset can sit on top of the other data.

ldecicco-USGS:
Changed to bug because it use to work, and now doesn't. I know when I tacked in the bgCol function initially, this example was great.

ldecicco-USGS:
(I'll be able to troubleshoot this afternoon)

ldecicco-USGS:
Just for posterity, the equivalent modern code is:

``` r
gs <- gsplot() %>%
  points( y=c(3,1,2), x=1:3, xlim=c(0,NA),ylim=c(0,NA),
          col="blue", pch=18, legend.name="Points", xlab="Index") %>%
  lines(c(3,4,3), c(2,4,6), legend.name="Lines", ylab="Data") %>%
[...]

PR #121, opened 2015-07-16
Fixes #83

### LINKED ISSUE #83: Legend should respect size

ldecicco-USGS:
Working in the airport...


jiwalker-usgs:
I think you might have not wanted to check in test.png?

ldecicco-USGS:
Where did that come from ?
one sec

R/axis.R
```diff
@@ -14,7 +14,7 @@
 #' @rdname axis
 #' @export
 #' @examples
-#' \dontrun{
```
jordansread, line 17: cheata gotta cheat

R/axis.R
```diff
@@ -27,7 +27,7 @@
 #'            col="blue", pch=18, legend.name="Points", xlab="Index") %>%
 #'    axis(side=c(3,4), labels=FALSE)
```
jordansread, line 28: how do you want this to work? 

``` R
 gs <- gsplot() %>%
    points(x=1:5, y=1:5, legend.name="Stuff", side=c(1,2,3,4)) %>%
    axis(side=c(3,4),labels=FALSE)
gs
```

not supported yet

jordansread, line 28: or 

``` R
 gs <- gsplot() %>%
    points(x=1:5, y=1:5, legend.name="Stuff", side=c(1,2)) %>%
    axis(side=c(3,4),labels=FALSE)
gs
```

which breaks now

ldecicco-USGS, line 28: Look here at how it use to look (ie...I want ticks with no labels on side 3 and 4)....I would think both strategies above should work (but I've ignored >2 sides so far...so that 1st example didn't use to work).

https://github.com/USGS-R/gsplot/issues/14

 (ignore the inset for now, the main plot has that sort of workflow as above)

jordansread:
we are expecting this to break the test. merging.

PR #139, opened 2015-07-22
Condense error_bar

ldecicco-USGS:
Got 'dumb' legend working (well, average-intelligence), and condensed error_bar functions to one. Still thinking about the best NA strategy there.


jordansread:
:+1:

PR #140, opened 2015-07-22
Moving par back to before plot.window

ldecicco-USGS:


ldecicco-USGS:
In this case:

``` r
gs <- gsplot() %>%
   points(y=c(3,1,2), x=1:3, xlim=c(0,NA),ylim=c(0,NA)) %>%
   error_bar(1:3, y=c(3,1,2), y.high=c(0.5,0.25,1), y.low=0.1) %>%
   error_bar(x=1:3, y=c(3,1,2), x.low=c(.2,NA,.2), x.high=.2, col="red",lwd=3)
gs
```

Would we want the horizontal `NA` error bar to have a vertical line, or no line?


 (moving to an issue in case this gets confusing)

jordansread:
@ldecicco-USGS I would say no line for `NA`

jordansread:
:+1:

PR #141, opened 2015-07-22
Allows limits to be stored as dates

### LINKED ISSUE #125: support date types
`graphics`:

``` R
plot(as.Date('1970-01-01'), 4)
```

image
`gsplot`:

``` R
gs = gsplot() %>% points(as.Date('1970-01-01'), 4)
gs
```

image

lindsayplatt:
Changes to `summarize_args` and ifelse statement in `set_view_lim` to allow class(xlim) to be Date. Only concern with line 169, accessing the x or y data points using var[1]. Need var = x, but var is [x, x1, x0].

Example:

``` r
dates <- seq(as.Date("1970-01-01"), as.Date("1970-01-31"), by="days")
y_vals <- runif(length(dates), min=0, max=100)

gs <- gsplot() %>% points(dates, y_vals, xlab="DATES", ylab="VALUES") # %>% axis(side=c(1,2), at=dates)
gs
```

Still does not completely fix #125 -- something with axes?

### DIFF
--- a/R/calc_views.R
+++ b/R/calc_views.R
@@ -129,7 +129,7 @@ set_view_lim <- function(views){
     match.side <- as.character(views_with_side(views, view.side))
     data.lim <- range(data[[var]][match.side],na.rm = T, na.action = NA)
     usr.lim <- views[[as.numeric(i)]][['window']][[lim.name]][1:2]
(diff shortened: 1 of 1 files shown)


ldecicco-USGS:
Wrong PR

ldecicco-USGS:
We'll need Dates/POSIXct/POSIXlt classes (so not just dates). Probably others. Let me think about this....

PR #156, opened 2015-07-23
Config from yaml

jiwalker-usgs:
Let me know what you think, seems to not break anything.


R/axis.R
```diff
@@ -14,12 +14,11 @@
 #' @rdname axis
 #' @export
 #' @examples
-#' 
 #' gs <- gsplot() %>%
 #'    points(x=1:5, y=1:5, legend.name="Stuff") %>%
 #'    lines(2:6, y=2:6, ylim=c(0,10)) %>%
 #'    axis(side=c(3,4),labels=FALSE) %>%
-#'    legend(location="topright")
+#'    legend("topright")
```
ldecicco-USGS, line 21: Why did you take that away? I think it's a coincidence that topright works correctly because that is the default in graphics, but we need to include `location` now so that we can also support default graphic legend. So, I'd rather keep all the locations in all examples.

R/axis.R
```diff
@@ -51,5 +50,8 @@ axis.gsplot <- function(object, ..., side=c(1,2)) {
 
 
   return(gsplot(object))
-
+
```
ldecicco-USGS, line 53: :+1:

ldecicco-USGS, line 53: Was just doing the same thing.

R/config.R
```diff
@@ -12,62 +12,33 @@
...
-    for(type in names(graphTemplate)[names(graphTemplate) %in% names(usrOptions)]){
-      graphTemplate[[type]] <- NULL
-    }
-    graphTemplate <- append(usrOptions, graphTemplate)
-  }
-  
-  options("gsplot"=graphTemplate)
-}
 
+  graphTemplate <- yaml.load_file(filename)
```
jordansread, line 22: exciting! :+1:

R/axis.R
jiwalker-usgs, line 21: This example somehow got screwed up in merge conflict resolution.  I thought I had everything from you, but not so.

ldecicco-USGS, line 21: OK cool, just curious if it was deliberate for some reason.

PR #159, opened 2015-07-25
Points.formula

jordansread:
got the other PR in here because I am bad :tophat: 

This works:

``` R
library(gsplot)
plot(-4:4, -4:4, type = "n") 
points(y~x, data=list(x=-3:3,y=-3:3)) 
# or
plot(-4:4, -4:4) 
points(y~x) 
```

 fails for this because it can't find limits:

``` R
gs <- gsplot() %>%
    points(y~x, data=list(x=-3:3,y=-3:3))
gs
```

but this works with a warning triggered by the same thing that makes the above fail:

``` R
gs <- gsplot() %>%
    points(y~x, data=list(x=-3:3,y=-3:3), ylim=c(-3,3),xlim=c(-3,3))
gs
```

The thing that fails is in calc_views, `set_view_lim(views)`, because it doesn't know to dive deeper into the list to get `x` and `y`. That might be a relatively simple fix.


jordansread:
Now this should work for 

``` R
gs <- gsplot() %>%
    points(y~x, data=list(x=-3:3,y=-3:3))
gs
```

PR #164, opened 2015-07-31
new set_args to deal with unnamed gsplot lists args

jordansread:
potential simpler pattern @ldecicco-USGS ?

Also, this catches the bug where unnamed vectors for x and y for `points` had undefined limits.


jordansread:
oh, it's the EXPLANATION title: 
image

ldecicco-USGS:
Huh

lindsayplatt:
that's one of the defaults I put in to follow USGS style guidelines

Issue #128, opened 2015-07-20
Embed callout + error_bar in point/line/


ldecicco-USGS:
Keep current functionality (condensing error_bar) but allow to be arguments in points/lines:

``` r
gs <- gsplot() %>%
    points(1:5, 1:5, 
                 callouts=list(labels=c(rep(NA, 4), "oh"))), 
                 error_bars=list(y.high=.1)))
gs
```

ldecicco-USGS:
@lindsaycarr 

I would think this:

``` r
gs <- gsplot() %>%
         lines(x=c(1,2), y=c(4,2), xlim=c(0, 5), ylim=c(0,5),
             callouts=list(labels=c(NA, "data"), col="blue"))
gs

```

Would only produce a `callout`at the 2nd data point, do you agree? So, we'll need to filter x/y to the callout/error bar calls.

lindsayplatt:
Yes, that would make way more sense. I'm working on changing it where the callouts function is called in points and lines

lindsayplatt:
@ldecicco-USGS  error_bar won't really work the same way though, there aren't labels defining where to put them or not. I would assume you would want these on all of the points if you used the embedded version. Otherwise just define them separately since you would have to enter the x/y values anyways?

ldecicco-USGS:
yeah, the standard `error_bar` function takes care of NA's, so I think within points/lines/etc it makes sense to do all-or-none

lindsayplatt:
Ohh, so you would be able to say 

``` r
gs <- gsplot() %>%
         lines(x=c(1,2), y=c(4,2), xlim=c(0, 5), ylim=c(0,5),
             error_bar=list(y.high=c(1, NA))
gs
```

to have an error bar for the first point but not the second.

ldecicco-USGS:
Yup.

lindsayplatt:
I didn't want to change the way NA's were handled in `points.R` where the embedded callouts function was called because the callouts functionality should be consistent whether it's embedded or not. So, I added to the actual `callouts.R` code (PR will come later today with some other commits)

So, these are the same:

``` r
gs <- gsplot() %>%
         points(x=c(1,2), y=c(4,2), xlim=c(0, 5), ylim=c(0,5)) %>%
         callouts(x=c(1,2), y=c(4,2), labels=c(NA, "data"), col="blue")
gs

gs <- gsplot() %>%
[...]

PR #167, opened 2015-08-02
mostly clean up and using set_args

jordansread:
added some documentation to 2 internal functions. using more consistent language (graphic_params -> function_args, as it isn't just for graphics package, and we call them args elsewhere). Should be mostly aesthetic change. 

Did add the option to specify something other than `.default` as the fall-back function for matching arguments. It isn't used, but could be fore something like `xy.coords`.


lindsayplatt:
@jread-usgs I was noticing that error_bar() function was not changed to use the new `set_args` function. Was that intentional? I'm close to finishing issue #128 but line 41 is causing a problem (https://github.com/USGS-R/gsplot/blob/master/R/error_bars.R#L41).

PR #173, opened 2015-08-03
is.gsplot() function

lindsayplatt:
Not integrated with `print` or anything - just works solo and returns a vector saying "Valid" or "Invalid" for each check


R/is_gsplot.R
```diff
@@ -0,0 +1,47 @@
...
+    class <- "Valid"	
+    
+    if (!"par" %in% duplicates) {
+      par_arguments <- "Valid"
+    } else { par_arguments <- "Invalid" }
+    
+  } else { class <- "Invalid" }
+  
+  validity <- rbind(class, par_arguments)
+  return(validity)
```
ldecicco-USGS, line 45: We'll want the return to be either TRUE or FALSE.

lindsayplatt, line 45: So one sweeping statement -- it's either all valid (TRUE) or at least one thing is invalid (FALSE)

ldecicco-USGS, line 45: Yup. These type of functions are usually used like (if(is.gsplot(x)){blah...)}

lindsayplatt, line 45: Ok, right - I've used `is.numeric` and stuff. I should have thought of that! Thanks :smile: new PR is in

Issue #157, opened 2015-07-25
need a is.gsplot() or something similar for gsplot class


jordansread:
would make testing etc easier. For example, gsplot can have any number of `points` or `lines` etc fields, but it should only have one `par` field. 

other checks could be put in here too.

ldecicco-USGS:
Expand `is.gsplot` (and I suppose create) to not just check the class but also if it's valid.

lindsayplatt:
What other checks will need to be in this function? I currently have it set up to check `class(object) == "gsplot"` and that there is only one `par`.

However, I cannot find a scenario where multiple `par` arguments will be in the object list. When I have added more `par()` statements, they are put under the same list argument:

``` r
usrDef <- gsplot() %>% 
    points(x=1, y=2, side=c(3,2), legend.name="Points 1", cex=3) %>% 
    points(x=3, y=4, side=c(1,4), legend.name="Points 2", pch=5, col="red") %>% 
    lines(x=c(3,4,3), y=c(2,4,6), legend.name="Lines 1", lty=5, col="orange") %>%
    lines(x=c(1,2,5), y=c(1,8,5), legend.name="Lines 2", lwd=3) %>%  
    lines(x=c(2,4), y=c(3,6), legend.name="Lines 3", lty=2, col="blue") %>%
    par(mfcol=c(2,1)) %>%
    par(las=2) %>%
    legend(x=1.5,y=4)
usrDef
```

lindsayplatt:
Additionally, should the function be internal (part of `print`) and halt plot execution? Current commit/PR just outputs vector saying "Valid" or "Invalid" for each check

ldecicco-USGS:
Definitely not internal. It is considered very good practice to have a validation function available.

lindsayplatt:
So this would just be for a user to check?

ldecicco-USGS:
A user or in the package...whereever

lindsayplatt:
Oh ok, I guess by internal I was thinking more along the lines of automated - somewhere the package would run this and if something was invalid, it would produce an error?

ldecicco-USGS:
I'm really not sure on how we are going to handle `par`. I can think of many situations where I actually would want multiple `pars`.....so we'll have to think about `par` in another conversation. As for other validity tests...we'll just have to start thinking about what's valid/not (maybe get inspiration in the tests). It shouldn't be too big of a function though. For now `class(object) == "gsplot"` is fine.

[...]

PR #174, opened 2015-08-03
is.gsplot() function

lindsayplatt:


lindsayplatt:
What is the "coverage" check?

ldecicco-USGS:
How well the overall package is being tests. The more unit tests we add, the better our coverage is. Other than that, @jread-usgs  is a fan - he'll have to answer, I don't use it much.

lindsayplatt:
Partially addresses #157

PR #175, opened 2015-08-03
Axes and ann up

ldecicco-USGS:
I learned an interesting thing about box()....taking it out screws up our plots. There is an argument in plots.default frame.panel which would be the one we want to use to decide if we want the plot boxed or not. When I just made that optional, plots started overplotting...so, we need to look into that.


R/print.R
```diff
@@ -48,16 +53,27 @@ print.gsplot <- function(x, ...){
...
+      mtext(text=window$xlab, side=window$side[1], line = 2)
+      mtext(text=window$ylab, side=window$side[2], line = 2)
+    }
     
     par(new=TRUE)
   }
-  box()
   
+  box()
+
```
ldecicco-USGS, line 80: I wanted box() to be conditional with "frame.plot", but that started to seriously screw up some plots. Overplotting all over the place. I really don't see what's special about box()...but soemthing to keep in mind.

PR #177, opened 2015-08-04
embedded error bars and callouts for points/lines + set_args for error_bar

lindsayplatt:


jordansread:
so the workflow for this is `points(x=1:2, y=3:4, callouts=list(labels=c('point1','point2'))`

would we consider this?

``` R
points(x=1:2, y=3:4, callouts(labels=c('point1','point2'))
```

which would be possible w/ using `substitute()` within the print method part of the `do.call()`?

lindsayplatt:
I think this is what we should aim for - it's way more intuitive, plus keeps the workflow from the callouts to embedded callouts the same. 

What exactly is `substitute()` doing?

jordansread:
I think the meat and potatoes are here: http://adv-r.had.co.nz/Computing-on-the-language.html

PR #184, opened 2015-08-06
symbols function and allows plot(5:7)

lindsayplatt:
These are the same:

``` r
#base R
plot(5:7)

#gsplot 
gs <- gsplot() %>%
     points(1:10)
gs
```


R/points.R
```diff
@@ -57,5 +57,12 @@ points.gsplot <- function(object, ..., legend.name=NULL, side=c(1,2)){
   to.gsplot <- list(list(arguments = do.call(set_args, c(fun.name, arguments)), 
                          gs.config=list(legend.name = legend.name, side = side))) %>% 
     setNames(fun.name)
+  
+  if (is.null(to.gsplot$points$arguments$y)){
+    to.gsplot$points$arguments$y <- to.gsplot$points$arguments$x
+    to.gsplot$points$arguments$x <- seq(length(to.gsplot$points$arguments$x))
+    to.gsplot$points$arguments$xlab <- "Index"
```
jordansread: want to make sure "Index" doesn't override what a user specified for `xlab` (if they did). You might want to check to see if `xlab` is in the `...` argnames

jordansread:
check on this? https://travis-ci.org/USGS-R/gsplot/builds/74470536#L480

R/points.R
lindsayplatt: Whoops I missed that, I've added a new line and will commit soon.

`if (is.null(to.gsplot$points$arguments$xlab)) to.gsplot$points$arguments$xlab <- "Index"`

lindsayplatt:
These checks are failing in [tests-points.R at line 52](https://github.com/USGS-R/gsplot/pull/159/files#diff-ea161b546703f4499de4b783b49e214eR52).

``` r
Warning messages:
1: In plot.xy(xy.coords(x, y), type = type, ...) :
  "formula" is not a graphical parameter
2: In plot.xy(xy.coords(x, y), type = type, ...) :
  "data" is not a graphical parameter
```

ldecicco-USGS:
So we gained the ability to have y=NULL, but lost the ability to do a formula? Let's spend a little time trying to work both in there. 

This might fall in the conversation @jiwalker-usgs and I were having yesterday. Do we want to let the user do 'dumb' things just because base R does? Like, specifying x when they mean y (`plot(x=1:10)` is actually `plot(y=seq_along(length(x), x=1:10)`)

lindsayplatt:
I just took another quick look and what we need is another check on the if statement that was added to allow `y=NULL`. When a formula is put in and it comes to that if statement, y is considered null because it doesn't exist. I added a check to the if statement to make sure no formulas were present AND y=NULL

lindsayplatt:
@ldecicco-USGS we don't need to have it label the x-axis with "index". I was just following with what base R did. I also thought that it might help people understand what is exactly happening when they only enter one set of numbers in.

ldecicco-USGS:
I think what you have is right (following base R's behavior). I'm just saying to extend base-R behavior, there's an argument "ann" that would allow the user to turn off the auto-labeling, and I think we'd need to incorporate that here. Don't worry about it for now because I think @jiwalker-usgs will make a case for deviating from base-R behavior on Tuesday.

R/points.R
```diff
@@ -57,5 +57,12 @@ points.gsplot <- function(object, ..., legend.name=NULL, side=c(1,2)){
   to.gsplot <- list(list(arguments = do.call(set_args, c(fun.name, arguments)), 
                          gs.config=list(legend.name = legend.name, side = side))) %>% 
     setNames(fun.name)
+  
+  if (all(names(to.gsplot$points$arguments) != "formula") && is.null(to.gsplot$points$arguments$y)){
```
ldecicco-USGS, line 61: We'll need to do this in (at least) `lines` too I'm guessing.

R/symbols.R
```diff
@@ -0,0 +1,58 @@
...
+symbols <- function(object, ...) {
+  override("graphics", "symbols", object, ...)
+}
+
+symbols.gsplot <- function(object, ..., legend.name=NULL, side=c(1,2)){
+  
+  current_list <- config("symbols")
+  arguments <- list(...)
+  
+  indicesToAdd <- !(names(current_list) %in% names(arguments))
```
jordansread, line 48: use `set_args` instead?

PR #187, opened 2015-08-10
curve function -- currently not working

lindsayplatt:
Desired workflow, but doesn't work:

``` r
gs <- gsplot() %>% 
    curve(x^2, from=1, to=5, col="blue")
gs

# Errors:
Error in curve.gsplot(object, ...) : object 'x' not found 
```

Bad workflow, but works with some warnings (however, incorrect "from-to" limits because of `add=TRUE`):

``` r
gs <- gsplot() %>%
    curve(expr=parse(text="x^2"), from=0, to=5, col="blue", add=TRUE)
gs
```

Same as before, except without `add=TRUE`, the y-axis label is jumbled but the limits are correct. 

``` r
gs <- gsplot() %>%
    curve(expr=parse(text="x^2"), from=0, to=5, col="blue")
gs
```


jordansread:
I will take a look at this.

jordansread:
this issue seems to be because `x^2` is evaluated with `list(...)`, as opposed to only being evaluated when the call is made to `graphics::curve` (i.e., in the print method). I think this is why ggplot2 uses lazyEval?

jordansread:
one way to deal with this is to do what R src does inside `curve`, which is to calculate an x and y, and then set up the actual call to (in our case) `lines`. This would break our current pattern, but is maybe worthy of discussion? 

something like this:

``` r
curve.gsplot <- function(object, ..., xname="x", legend.name=NULL, side=c(1,2)){

  current_list <- config("curve")


  arguments <- ??convert to lines arguments by doing ?? 
   # x <- seq.int(from, to, length.out = n)
   # ll <- list(x = x)
   # names(ll) <- xname
   # y <- eval(expr, envir = ll, enclos = parent.frame())
  # prefer that we use `set_args` here?:
  indicesToAdd <- !(names(current_list) %in% names(arguments))
  arguments <- append(arguments, current_list[indicesToAdd])

  object <- append(object,  list(lines = list(arguments = arguments, 
                                              gs.config=list(legend.name = legend.name, 
                                                             side = side))))

  return(gsplot(object))

}
```

jordansread:
also, not sure why `xname` has a default, but none of the other args do. I think the method signature should be:

``` R
curve.gsplot <- function(object, ..., legend.name=NULL, side=c(1,2)){
}
```

lindsayplatt:
Oh I think the xname is leftover from one of the times I was messing around, but it made no difference and I forgot to delete it. 

I had also considered calculating the x and y first - it seems like the quickest/easiest fix. 

Also, I keep forgetting that `set_args` is a thing now - I will make sure to add it to the final product of this and to the `symbols` function that was just merged.

jordansread:
Ok, sounds good. I think the key here to avoid that error is to use `substitute(...)` and then some other magic.

lindsayplatt:
Using `args = as.list(substitute(list(...)))[-1L]` that you suggested in #188 and adding it to do.call in `calc_views`, the code gets hung up in override now because the object is coming in as `x^2` from L83 of print. Making progress though :smiley: 

`Error in x^2 : non-numeric argument to binary operator`

jordansread:
yep, this is just one piece of the pie. I still think that for curve, you will need to create a `lines` list to append to gsplot, instead of relying on calling curve later. Calling it later will screw up things like limits etc.

lindsayplatt:
@ldecicco-USGS and @jread-usgs -- curve should be working now. I've also added commits addressing the use of `set_args` and having `lines(1:10)` work

Issue #185, opened 2015-08-10
error_bar, callout, axis limits


ldecicco-USGS:
limits need to be incorporated into our smart limits.

jordansread:
Can we get a reproducible example (code + screenshot) of this bug?

ldecicco-USGS:
``` r
gs <- gsplot() %>%
   points(1:10, 1:10) %>%
   error_bar(5, 5, y.high=50) 
gs 
```

ldecicco-USGS:
So I think just moving the logic from error_bar.default to error_bar.gsplot could take care of this. Same with callouts.

lindsayplatt:
What about in the calc_views `set_view_lims` function somewhere? Then we wouldn't have to change to callouts and error_bars, just add a line or two checking for those before defining the limits

ldecicco-USGS:
Either way would work. The question is which function do we want to muck up the most. I like the idea of keeping the calc_view stuff as generalized as possible and leaving all the function specific stuff in the  functions.

lindsayplatt:
Ah ok, good point

jordansread:
I think I know what we are talking about. If so, I think the logic should be in error_bar.gsplot..I think. (and callouts)

lindsayplatt:
This is how the limits automatically look for arrows -- you can't see the horizontal bar ending the line because it is on the axis

![image](https://cloud.githubusercontent.com/assets/13220910/9262214/cb19fa10-41d8-11e5-85c7-049268b9aed6.png)

ldecicco-USGS:
I don't think that's an issue. We butt up points and lines to the tippy top already. An issues is that we should be letting the user pick yaxs and xaxs, but I don't think we are doing this right now.

From `?par`:
The style of axis interval calculation to be used for the x-axis. Possible values are "r", "i", "e", "s", "d". The styles are generally controlled by the range of data or xlim, if given.
Style "r" (regular) first extends the data range by 4 percent at each end and then finds an axis with pretty labels that fits within the extended range.
Style "i" (internal) just finds an axis with pretty labels that fits within the original data range.
Style "s" (standard) finds an axis with pretty labels within which the original data range fits.
Style "e" (extended) is like style "s", except that it is also ensures that there is room for plotting symbols within the bounding box.
Style "d" (direct) specifies that the current axis should be used on subsequent plots.
(Only "r" and "i" styles have been implemented in R.)

ldecicco-USGS:
[...]

PR #202, opened 2015-08-13
pt.bg color in legend + error_bar calculated into limits

lindsayplatt:


lindsayplatt:
Embedded error_bars will not work now and I think I know why -- just can't get to it before tomorrow. 

Also, the horizontal segment at the end of the error_bars is covered by the axes (see my image on issue #185). Is this an issue?

I just want to get these fixes in for @mhines-usgs, and will continuing tweaking tomorrow morning.

inst/extdata/default.yaml
```diff
@@ -1,5 +1,5 @@
-xaxs: "i"
-yaxs: "i"
+xaxs: "r"
```
ldecicco-USGS, line 1: Let's keep it in so that we can get the bugs on the user specified limits fixed. Obviously our yaml is going to change a lot before we introduce the package to the wild, and picking our favorite defaults might take some back-and-forth discussions.

lindsayplatt, line 1: Do you mean revert this back to "i" for the time being?

ldecicco-USGS, line 1: I was going to say that and changed my mind. Everyone should know though that all of our examples will now look different. I think we leave it as "r" and bump up the priority to fix the user-specified limit stuff.

lindsayplatt, line 1: Ok, I will assign that to myself and make sure it gets done next week.

ldecicco-USGS, line 1: Awesome!

PR #217, opened 2015-08-25
vignette populated with Maumee data examples

lindsayplatt:


ldecicco-USGS:
Run:
`devtools::build_vignettes()`
to add the R/html/Rmd to the inst folder. Then we can look at the pretty html in rawgit

ldecicco-USGS:
See:

https://rawgit.com/lindsaycarr/gsplot/master/inst/doc/gsplotIntro.html

lindsayplatt:
cool!

ldecicco-USGS:
I'd vote we 'Suggest' `dplyr` so we don't have to do anything like:

``` r
sapply(sites, function(x) MaumeeDV$Date[which(MaumeeDV$site_no==x)], USE.NAMES=TRUE)
```

@jread-usgs thoughts?

ldecicco-USGS:
The vignette nicely shows that the grid is sucking again...wtf. I'm going to merge it, and we can iterate on what we want in there.

lindsayplatt:
yeah there is also a problem with where the legend is placed when it is outside the plot (see Fig. 4)

ldecicco-USGS:
Also, I'd 'Suggest' `captioner` so we don't have to manually number the figures. I'd prefer `kfigr`....but there was an issue with it the last time I tried it in a vignette. It might be fixed now. One of those would be good if we want to number (and reference if using `kfigr`) the figures.

Issue #211, opened 2015-08-24
add new type - xaxs/yaxs='o' 


lindsayplatt:
's' specifies single buffer (check that 's' isn't already used)

``` r
gs <- gsplot() %>%
   points(1:2, 1:2, ylim=c(0,NA), yaxs='s')
gs
```

does not buffer zero, but adds buffer to upper limit (4%)

lindsayplatt:
's' is already a type 

```
xaxs
The style of axis interval calculation to be used for the x-axis. Possible values are "r", "i", "e", "s", "d". The styles are generally controlled by the range of data or xlim, if given.
Style "r" (regular) first extends the data range by 4 percent at each end and then finds an axis with pretty labels that fits within the extended range.
Style "i" (internal) just finds an axis with pretty labels that fits within the original data range.
Style "s" (standard) finds an axis with pretty labels within which the original data range fits.
Style "e" (extended) is like style "s", except that it is also ensures that there is room for plotting symbols within the bounding box.
Style "d" (direct) specifies that the current axis should be used on subsequent plots.
(Only "r" and "i" styles have been implemented in R.)
```

lindsayplatt:
going to use `Style "o" (one)` to specify only one side gets buffered

jordansread:
probably a good plan. Note that 's' isn't implemented in R (is a carryover from S), but it is probably reserved in some way for the future and we probably should try to avoid conflicts. 'o' works.

lindsayplatt:
Where should we document this feature? I was thinking in the gpsplot par?

lindsayplatt:
I think the overhaul for iteratively building the gsplot object has impacted this. A test was created but this hasn't been fixed since the changes, so the test has been failing.

lindsayplatt:
The `to.gsplot` line in `set_args` puts `yaxs='o'` into `window$par`, rather than as an argument of points. Therefore, it is ignored in `summarize_args` in calc_views>set_view_lim. See #276

jordansread:
We need to re-implement this w/ the change to including `sides` as a top level field

PR #221, opened 2015-09-01
yaxs/xaxs='o' option

lindsayplatt:
addresses #211 

gs <- gsplot()  %>%
    points(1:4, xlim=c(NA,10), xaxs='o')  %>%
    lines(1:6)
gs


R/calc_views.R
```diff
@@ -135,6 +137,18 @@ set_view_lim <- function(views){
       usr.lim <- views[[as.numeric(i)]][['window']][[lim.name]][1:2]
       views[[as.numeric(i)]][['window']][[lim.name]] <- data.lim
       views[[as.numeric(i)]][['window']][[lim.name]][!is.na(usr.lim)] <- usr.lim[!is.na(usr.lim)]
+      
+      usr.axs <- data[[paste0(var, 'axs')]][[as.numeric(i)]]
+      view.index <- which(!names(views[[as.numeric(i)]]) %in% c('window', 'gsplot'))[which(!is.na(usr.axs))]
+      
+      if (any(!is.na(usr.axs)) && usr.axs[which(!is.na(usr.axs))] == 'o') {
```
lindsayplatt, line 145: @ldecicco-USGS @jread-usgs can you review this and help me make it less messy? There's so much indexing/subsetting craziness (lines 145-150 in particular)

jordansread, line 145: I will take a look

R/calc_views.R
```diff
@@ -107,7 +107,9 @@ set_view_lim <- function(views){
   
 
   data <- list(y=summarize_args(views,c('y','y1','y0'),ignore=c('window','gs.config')), 
-               x=summarize_args(views,c('x','x1','x0'),ignore=c('window','gs.config')))
+               x=summarize_args(views,c('x','x1','x0'),ignore=c('window','gs.config')),
+               yaxs=summarize_args(views,c('yaxs'),ignore=c('window','gs.config')),
```
jordansread: I would call `x` and `y` data, but the `yaxs` and `xaxs` probably shouldn't be called that (just for readability's sake)

lindsayplatt: ah - ok, I'll put those in a separate list

jordansread:
also, just a note. I know it is confusing, but we are trying to be consistent w/ styles. So `camelCase` is to be avoided, even though there are tons of examples that use it within the package.

R/calc_views.R
jordansread: I've got a PR coming to your branch. Another reason to avoid that is because you loop through all items in `data` on line 118, and you don't want to loop the `yaxs` right?

lindsayplatt:
Ok, I can't seem to find where I added that style of variable naming. Can you point it out? 

Also, very useful to know, I have been unsure about which is preferred.

R/calc_views.R
```diff
@@ -135,6 +137,18 @@ set_view_lim <- function(views){
...
+      usr.axs <- data[[paste0(var, 'axs')]][[as.numeric(i)]]
+      view.index <- which(!names(views[[as.numeric(i)]]) %in% c('window', 'gsplot'))[which(!is.na(usr.axs))]
+      
+      if (any(!is.na(usr.axs)) && usr.axs[which(!is.na(usr.axs))] == 'o') {
+        buffer <- 0.04*views[[as.numeric(i)]][['window']][[lim.name]][[2]] #4% buffer based on upper limit
+        buffer.lim <- views[[as.numeric(i)]][['window']][[lim.name]][[which(is.na(usr.lim))]]
+        buffer.lim <- ifelse(which(is.na(usr.lim)) == 1, buffer.lim - buffer, buffer.lim + buffer)
+        views[[as.numeric(i)]][['window']][[lim.name]][[which(is.na(usr.lim))]] <- buffer.lim
+        views[[as.numeric(i)]][[view.index]][[paste0(var, 'axs')]] <- NULL
+        views[['par']][[paste0(var, 'axs')]] <- 'i'
```
jordansread: can you explain what is going on here? (setting par(xaxs) = 'i')?

jordansread:
oh, I guess you didn't add any camelCase, sorry.

R/calc_views.R
lindsayplatt: The `default.yaml` has `yaxs = 'r'` and `xaxs = 'r'`, so when the "o" style is selected we add the 4% buffer to one side and so will want the axes to have the 'i' style so that they have the exact limits we defined.

R/calc_views.R
```diff
@@ -135,6 +137,18 @@ set_view_lim <- function(views){
       usr.lim <- views[[as.numeric(i)]][['window']][[lim.name]][1:2]
       views[[as.numeric(i)]][['window']][[lim.name]] <- data.lim
       views[[as.numeric(i)]][['window']][[lim.name]][!is.na(usr.lim)] <- usr.lim[!is.na(usr.lim)]
+      
+      usr.axs <- data[[paste0(var, 'axs')]][[as.numeric(i)]]
+      view.index <- which(!names(views[[as.numeric(i)]]) %in% c('window', 'gsplot'))[which(!is.na(usr.axs))]
+      
+      if (any(!is.na(usr.axs)) && usr.axs[which(!is.na(usr.axs))] == 'o') {
+        buffer <- 0.04*views[[as.numeric(i)]][['window']][[lim.name]][[2]] #4% buffer based on upper limit
```
jordansread: does this work when the axes is reversed? How about when it is logged?

jordansread: Also, I don't think this calculation mirrors the one done in R for buffers. you might want to take a look at that (I think it uses range or diff for the 4% and not just the top lim)

R/calc_views.R
lindsayplatt: It's actually not looping through the `yaxs`, because of line 117, which filters only the x and y. But I agree that it should be in a separate list anyways

jordansread:
One way this can be cleaner is to separate the `user_lims` logic from the `data_lims`, perhaps as two separate functions that can both be tested against the different possible inputs/outputs.

R/calc_views.R
lindsayplatt: It does work for reversed axes because the reversal happens after the window limits are set:

``` r
gs <- gsplot()  %>%
    points(1:4, xlim=c(0,NA), xaxs='o')  %>%
    axis(side=1, reverse=TRUE)
gs
```

logged axes do not work (thanks for pointing that out)

jordansread: ok, good to know. The logging is a challenge because it is handled within plot.window the way we have `print` set up now

lindsayplatt: You're right - it does use the range, it just so happened that on the example I used, it worked out with the upper limit

lindsayplatt:
so two functions used within the `set_view_lim` function?

R/calc_views.R
```diff
@@ -135,6 +137,18 @@ set_view_lim <- function(views){
       usr.lim <- views[[as.numeric(i)]][['window']][[lim.name]][1:2]
       views[[as.numeric(i)]][['window']][[lim.name]] <- data.lim
       views[[as.numeric(i)]][['window']][[lim.name]][!is.na(usr.lim)] <- usr.lim[!is.na(usr.lim)]
+      
+      usr.axs <- data[[paste0(var, 'axs')]][[as.numeric(i)]]
+      view.index <- which(!names(views[[as.numeric(i)]]) %in% c('window', 'gsplot'))[which(!is.na(usr.axs))]
```
jordansread: should this be within the `if` block, since it is only relevant/used there?

lindsayplatt: yes, another thing I missed

jordansread:
There is also a nasty error when a user tries:

``` R
gs <- gsplot() %>%
    points(1:4, xlim=c(1,3), xaxs='o') %>%
    lines(1:6)
gs
```

lindsayplatt:
that should throw an error because you aren't indicating which limit should be buffered and which shouldn't

jordansread:
yeah, you might want to give it better text for the error to throw and check for it.

PR #222, opened 2015-09-01
last additions for yaxs/xaxs = "o"

lindsayplatt:
addresses more of #211

### DIFF
--- a/R/calc_views.R
+++ b/R/calc_views.R
@@ -138,10 +138,8 @@ set_view_lim <- function(views){
       data.lim <- range(data.var[is.finite(data.var)])
       usr.lim <- views[[n.i]][['window']][[lim.name]][1:2]
       views[[n.i]][['window']][[lim.name]] <- data.lim
-      
-      if (any(!is.na(usr.lim))){
-        views[[n.i]][['window']][[lim.name]][!is.na(usr.lim)] <- usr.lim[!is.na(usr.lim)]
-      }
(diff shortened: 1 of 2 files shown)


lindsayplatt:
The logged axis issue is separate from the yaxs="o" (buffer is calculated the same way as non-log scale limits). The error we were seeing is from trying to set the limit as 0 (see #223 for discussion on whether we should have a fix for this). 

A logged axis will work if you don't try to use 0 as one of the limits:

``` r
y <- runif(10,1,40000)
x <- 1:10

gs <- gsplot()  %>%
    points(x,y, ylim=c(10,NA), log='y') 
gs
```

lindsayplatt:
@jread-usgs this should be good to go now (don't want to merge without other eyes on it first)

jordansread:
hmm...did you try knitting the README.Rmd? I did and got some changes to the plots (like the second one attached here). 

image

lindsayplatt:
weird, I just ran that in my environment and got: 

``` r
gs <- gsplot() %>%
  points(y=c(3,1,2,4,5), x=c(1:3,8,80), col="blue", pch=18, legend.name="Points") %>%
  lines(c(3,4,3), c(2,4,6), legend.name="Lines", ylab="logged y axis", xlab="logged x axis", log='xy') %>%
  callouts(x=8, y=4, lwd=2, angle=45, labels="Not sure about this one") %>%
  title("logged axes") %>%
  axis(side=c(1,2,3,4), labels=FALSE, n.minor=4)
gs
```

image

lindsayplatt:
It's weird that the labels are stopping at 50 though, right?

Issue #193, opened 2015-08-12
should we be building the plotting object iteratively instead of all at once?


jordansread:
curious of views and other things should be built as you add to the gsplot object, or if they should be done all at once as part of `print`. 

breaking it up would help us w/ development debugging. Not sure how I feel on the topic, but since we are talking about modifying the render object structure I thought I would bring this up.

jordansread:
also, one advantage of this is that if we have conflicting things specified in the views, the graceful failover is just to use the most recent setting. So if you had

``` R
gs <- gsplot() %>%
  points(1:2,1:2,xlab='dog') %>%
  lines(3:4, 5:6, xlab='cat')
gs
```

the xlab would be cat. More advantageous for more complicated examples, but you get the point.

ldecicco-USGS:
Interesting...might be helpful to people using it debugging too.

Edit: I can see myself saying..."OHHH that's why the limit's all screwed up"

jiwalker-usgs:
I was thinking about this as well, and I think it is reasonable as long as there isn't too much to deal with as far as maintaining the integrity of the object.  I've been writing this comment for a while and trying to come up with drawbacks, but can't come up with anything specific.  If you want to discuss, I probably have some strong opinions one way or another.

jordansread:
are we doing this?

ldecicco-USGS:
It's my vote.

jiwalker-usgs:
I think I'm convinced.

jordansread:
I am trying this

jiwalker-usgs:
:+1:

jordansread:
WIP here: https://github.com/jread-usgs/gsplot/tree/iterative-builder

but I broke legend and I think I broke logged axes when using multiple calls to the same side

jordansread:
that PR is ready for review

Issue #224, opened 2015-09-02
set_view_order doesn't work as it should. pulls view args up into top level


jordansread:
Noticed this when working on #193 

``` R
gs <- gsplot()  %>%
  points(1:10, col="black")  %>%
  bgCol(col="coral2")
#then putting a breakpoint on the return of set_view_order:
names(append(dataViews, otherViews))
[1] "view"   "window" "points"
#when points shouldn't be moved up to high level (window either). 
```

ldecicco-USGS:
So what's going on here?

```
gs <- gsplot()  %>%
     points(1:10)
gs
```

We lost the ability to only specify 1 vector it seems.

jordansread:
@ldecicco-USGS see #245. oh yes, you saw it

jordansread:
no, that example is an old version, but the behavior on v 0.3+ is the same

jordansread:
ignore the y=NULL issue, that is different

lindsayplatt:
@jread-usgs @ldecicco-USGS I don't think this issue is relevant anymore. Can someone look and then close if you agree?

lindsayplatt:
There appear to be two issues mentioned here: 
1. Points being top-level. This is no longer an issue:
```r
gs <- gsplot()  %>%
  points(1:10, col="black")  %>%
  bgCol(col="coral2")
names(gs)
[1] "side.1"   "side.2"   "global"   "view.1.2"
```
2. Losing the ability to only specify one vector of points, which is being addressed in #245 .

@ldecicco-USGS and @jread-usgs please close if you agree that this issue is no longer relevant.

PR #225, opened 2015-09-02
Iterative builder

jordansread:
is a WIP for #193


jordansread:
I could use a review if others are up for it. 

I would like to have legends iteratively build too, but that code is kinda huge right now to tackle. I punted and added the legend name to each call, then ignore it in print. Kind of breaks the pattern, but it was a way around a gigantic overhaul of `legend`

R/print.R
```diff
@@ -71,10 +71,8 @@ print.gsplot <- function(x, ...){
...
   }
 
-  draw_legend(x)
-  draw_axis(x)
-
-  # par(defaultPar)
+  draw_legend(views)
+  draw_axis(views)
   
 }
```
ldecicco-USGS, line 78: I can't seem to write a comment next to a picture...but it's interesting how the ticks are behaving now. The top ones cover the legend, the left don't

jordansread, line 78: see this line: https://github.com/USGS-R/gsplot/pull/225/files#diff-832bc7f54d2e9d5191b3bdcef9979848R20

I think that is why

jordansread, line 78: and https://github.com/USGS-R/gsplot/issues/224

PR #227, opened 2015-09-03
multiple callouts not broken

lindsayplatt:
Only way I could think to remedy the issue was by using a for loop. Using an apply function wasn't making sense to me, but others might have a better idea. Also, I had to add some initialization of objects before the loop in order for it to grab and store the correct values. 

PT brought up an issue with multiple callouts requested:

``` r
gs <- gsplot()  %>% 
    points(c(1,2), c(2,3))  %>% 
    callouts(c(1,2), c(2,3), labels=c("more", "stuff"))
gs
```

The if statements used for `angle=auto` did not understand how to handle multiple xy points.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: gsplot
 Type: Package
 Title: Geological Survey Plotting
-Version: 0.2.2
+Version: 0.2.3
 Date: 2015-07-22
 Authors@R: c( person("Jordan", "Read", role = "aut",
     email = "jread@usgs.gov"),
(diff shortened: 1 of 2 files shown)


lindsayplatt:
@jread-usgs @ldecicco-USGS that vignette fail wasn't due to the dates. It was because a single callout was no longer working due to this apply function: `apply(good.pt, 1, function(z){`. A single callout would have dim = NULL, so I added an if statement to catch those conditions

lindsayplatt:
I would like to merge this once the checks pass so PT can have the graphs look nice again

ldecicco-USGS:
Yup!

PR #228, opened 2015-09-04
Iterative builder

lindsayplatt:

PR #231, opened 2015-09-09
changes recommended by #228 code review

jordansread:
from #228


jordansread:
@lindsaycarr or @ldecicco-USGS want to take a quick look and merge if appropriate? These are all based on comments from our review

ldecicco-USGS:
Will do.

PR #235, opened 2015-09-10
Embed function

jordansread:
anyone interested can take a look:

``` R
gs <- points(gsplot(), c(0,3), c(2,4),  
               error_bar(x.low=c(NA,1))) 
```

image

and callouts too


jiwalker-usgs:
cool

lindsayplatt:
:+1: much more intuitive than having to also say `list()` within the embedded feature. I don't know if you're finished or not, but don't forget that the `lines` function will need the same overhaul that you did to points!

ldecicco-USGS:
This makes me happy!

jordansread:
note that this currently doesn't support multiple embedded functions, but I see several use cases for that:

``` R
gs <- lines(gsplot(), c(0,3), c(2,4), callouts(labels=c('dogs','cats'),col='yellow'), error_bar(x.low=c(NA,1)))
```

upvote if you want this?

jiwalker-usgs:
:+1: it will likely be non-intuitive and hard to deal with error if users can't do two embedded calls.  Also, we may want some fanciness in the future that would make this use case more common.

ldecicco-USGS:
:+1: Upvoting. I don't have time today for a review, but if you feel like merging, I can live with discovering issues as we go (on this!)

Issue #246, opened 2015-09-16
par and limit extractor functions


ldecicco-USGS:
```
gs <- gsplot() %>%
  points( y=c(3,1,2), x=1:3, xlim=c(0,NA),ylim=c(0,NA),
          col="blue", pch=18, legend.name="Points", xlab="Index") %>%
  lines(c(3,4,3), c(2,4,6), legend.name="Lines", ylab="Data") %>%
  abline(b=1, a=0, legend.name="1:1") %>%
  legend(location="topleft",title="Awesome!") %>%
  grid() %>%
  bgCol(col="lightgrey") %>%
  error_bar(x=1:3, y=c(3,1,2), y.high=c(0.5,0.25,1), y.low=0.1) %>%
  arrows(x0=0.75, y0=2, x1=1, y1=2.8, lwd=2) %>%
  title("Graphing Fun") %>%
  text(.75,1.75,labels="Weird data") 
gs
u <- par("plt")
v <- c( (u[1]+u[2])/2, u[2], (u[3]+u[4])/2, u[4] )
config("par",new=TRUE,fig=v,mar=c(1,1,1,1))
gs2 <- gsplot() %>%
  points(1, 2) %>%
  bgCol(col="white")
gs2
# To get back to the 'default par':
par(gs.par(gs))
```

Suggestions?

ldecicco-USGS:
ie. `par.gs`

ldecicco-USGS:
Others?
xlim.gs
ylim.gs
par.gs

jordansread:
`log.gs`

jordansread:
Anyone started on this? If not, I might take a shot. Flight is delayed :airplane:

ldecicco-USGS:
Of course it is. Do it. These will be uber-handy functions.

jordansread:
:airplane: ORD :airplane: ery-time

lindsayplatt:
these will be super handy for some repgen features too

jordansread:
``` R
 usrDef <- gsplot(mar=c(4,4,4,4), xaxs='r', yaxs='r') %>% 
    points(x=1, y=2, side=c(3,2), legend.name="Points 1", cex=3, xlab='cat')

xlim(usrDef)[[1]]
c(1,1)

  usrDef <- points(usrDef,x=3:10,y=4:11, side=c(3,2))

xlim(usrDef)[[1]]
c(1,10)
```

jordansread:
``` r
[...]

PR #251, opened 2015-09-16
view-specific par args

jordansread:
#246


jordansread:
whoops. Looks like I have work to do on the README images. consider this a WIP but comments welcome. 

This kind of thing is supported now:

``` R
points(gsplot(), y=1, x=2, col="blue", pch=18, las=3)  %>% lines(2:3,4:5)  %>% points(3,4,side=c(3,4),las=1)
```

image

R/calc_views.R
```diff
@@ -61,6 +62,13 @@ group_views <- function(gsplot){
   return(append(views, non.views))
 }
 
+append_replace <- function(old.list, new.list){
+  out.list <- old.list
+  out.list[names(old.list) %in% names(new.list)] <- new.list[names(out.list[names(old.list) %in% names(new.list)])]
+  new.list[names(new.list) %in% names(old.list)] <- NULL
+  out.list <- append(out.list, new.list)
+  return(out.list)
+}
```
ldecicco-USGS, line 71: So can we get rid of set_window?

jordansread, line 71: working on that now...it still does a few important things, but I am trying to figure out what they are.

jordansread, line 71: I think i need to keep `set_window` in there, but we should be able to get rid of all or most of it in the future. I just don't really have a good grasp of what it does, and playing around w/ it breaks things.

jordansread:
OK, this is ready for review now

R/set_args.R
```diff
@@ -26,7 +26,8 @@ set_window_args <- function(object, fun.name, ..., legend.name=NULL, side=c(1,2)
   else
     arguments = set_args(fun.name, package=package)
   e.fun = dots$e.fun
-  to.gsplot <- list(list(arguments = arguments, gs.config=list(legend.name = legend.name, side = side))) %>% 
+  to.gsplot <- list(list(arguments = append(formal_arguments(arguments, def.funs, names(config(fun.name))), window_arguments(arguments, def.funs)),
+                         gs.config=list(legend.name = legend.name, side = side, par=par_arguments(arguments, def.funs)))) %>%
```
lindsayplatt, line 30: see #267 (for `text()`, y is not a formal argument and gets dropped here)

jordansread, line 30: in the call, need to do `def.funs=c('graphics::text','graphics::text.default')` I think (in text.R)

Issue #109, opened 2015-07-15
multi-gsplots don't work with layout()


jordansread:
guessing it has something to do w/ https://github.com/USGS-R/gsplot/blob/master/R/print.R#L24

jordansread:
see @lindsaycarr 's example on #94

jordansread:
``` R
layout(c(1,2,3,4))
plot(1,2)
dev.cur()
[1] 2
```

which is why we don't get a plot.new() when we use layout.

jordansread:
played around with this for too long. setting to wontfix and we should just push people to use split.screen(). Not sure if par(mfcol=2) works or not.

ldecicco-USGS:
I think jumping to 'won't fix' is a bit of a rash decision....is there a 'table issue for now' option. In an ideal world this works.

thongsav-usgs:
This will likely be an immediate impact on repgen's UV hydrographs, which currently uses layout and has N plots.

lindsayplatt:
I can spend some time working on this @jread-usgs

lindsayplatt:
The same thing happens with split.screen, so the issue must be somewhere else. 

``` r
split.screen(figs = c(2,1))
dev.cur()
```

jordansread:
@ldecicco-USGS `wontfix` for right now, because we can use split.screen(). I will create a longer term label. Just didn't want to dwell on this because split.screen will serve near-term issues.

@lindsaycarr  doesn't 

``` R
split.screen(c(1,2))
gs
gs
```

work for you?

ldecicco-USGS:
This works now:

``` r
library(RColorBrewer)
num.plots <- 4
ncol <- 2
nrow <- 2
plotmatrix <- matrix(seq(1:num.plots), ncol, nrow) 
colors <- brewer.pal(length(plotmatrix), "Reds")

png("test.png")
layout(plotmatrix)
for(i in 1:num.plots){
  gs<-gsplot() %>%
    points(i,2,col=colors[i],pch=19,cex=2) 
  print(gs)
}
dev.off()
```

![test](https://cloud.githubusercontent.com/assets/1105215/8830646/f6f80816-3062-11e5-891e-f215332dde65.png)

aappling-usgs:
This seems to be a problem again...

@ldecicco-USGS's example above doesn't work for me:

``` r
library(RColorBrewer)
num.plots <- 4
ncol <- 2
nrow <- 2
[...]

PR #259, opened 2015-10-06
Result of commenting out par(top.par).

ldecicco-USGS:


inst/doc/gsplotIntro.Rmd
```diff
@@ -114,7 +114,7 @@ For timeseries, it is sometimes helpful to plot data on the same graph to make c
 
 ```{r echo=TRUE, fig.cap="Fig. 4 Water temperature timeseries on primary y-axis with pH timeseries on secondary y-axis.", fig.width=6, fig.height=6}
 site <- '04193490'
-demoPlot <- gsplot(mar=c(7.1, 4.1, 4.1, 2.1)) %>% 
+demoPlot <- gsplot(mar=c(7.1, 4.1, 4.1, 4.1)) %>%
```
ldecicco-USGS, line 117: This plot still shows that if 2 views share an axis, the axis call is being repeated. So, doesn't actual solve #248

ldecicco-USGS:
Fixes #109

R/print.R
```diff
@@ -42,7 +42,7 @@ print.gsplot <- function(x, ...){
 
   top.par <- par(no.readonly=T)
   for (i in which(names(views) %in% 'view')){
-    par(top.par)
+    # par(top.par)
```
jordansread, line 45: I don't understand why this was the issue w/ layout. This is a loop within 1 gsplot object. layout was calling 3 or 4 different gsplot objects.

ldecicco-USGS:
Might be more to it, but my initial guess is that layout does some magic to change par$usr

R/print.R
ldecicco-USGS, line 45: Wrote this in the wrong place. I think it's because the layout function knows how to define par$usr

jordansread:
Just so I understand this, the `par(top.par)` literally did nothing other than break `layout`? I worry that it was doing something that we don't actually test for...but I could be wrong.

ldecicco-USGS:
That's what I worry about too. As far as I know, and as far as what's being tested for, I haven't found anything that is now broke. But if anything jiggles your memory as to why it's in there, now's the time to think of it. I'll quick run through all the examples by hand and make sure they look like I think they should look.

jordansread:
:+1:

PR #264, opened 2015-10-07
enhancements for gsplot (not actually to merge) 

lindsayplatt:
I created this PR from a new branch so that I could upload the .Rmd file I put together (.Rmd is not a file type supported to include in an issue) and we could have a thread about the graphs that should or shouldn't be added. The file in this PR has a table of contents showing which I think should be added (ADD), which gsplot can already handle (GSPLOT), and which we most likely do not want to be in gpslot (DON'T NEED TO ADD). We don't need to add plot types that just do data manipulation (e.g. qqplot) and then still give lines and points; I tried to only say "ADD" for plots that gsplot literally cannot create right now (e.g. boxplots, histograms). 

@ldecicco-USGS I did not have time to put a yaml together, but wanted to get this out there and get discussion going. 

**The gsplot version of the report graph was only rendering using gsplot v0.2.3 for some reason.

### DIFF
--- a/inst/doc/gsplot_enhancements.Rmd
+++ b/inst/doc/gsplot_enhancements.Rmd
@@ -1,0 +1,614 @@
+---
+title: "gsplotEnhancements"
+author: "Lindsay R. Carr"
+date: "October 6, 2015"
+output:
+  html_document:
+    toc: yes
+  pdf_document:
+    toc: yes
+---
+
+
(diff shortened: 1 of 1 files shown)


jordansread:
@lindsaycarr what should we do w/ this PR? Turn it into an issue? Add to wiki?

lindsayplatt:
@jread-usgs I guess make it an issue? I can just upload the .Rmd file

Issue #239, opened 2015-09-11
title ends up in a view, not as top level


jordansread:
Should be top level, like par, legend, etc.

jordansread:
from UVHydro:

``` R
uvhplot <- axis(uvhplot, side=2) %>% 
    axis(side=1,at=primaryInfo$dates,labels=as.character(1:length(primaryInfo$dates))) %>%
    grid(nx=0, ny=NULL, equilogs=FALSE, lty=3, col="gray") %>% 
    abline(v=primaryInfo$dates, lty=3, col="gray") %>% 
    legend(location="below", title="") %>%
    title(main="", xlab=primaryInfo$date_lbl, ylab=primaryInfo$primary_lbl) 
```

ldecicco-USGS:
`grid` and `bgCol` too I think would make sense. Then, we don't have to deal with ordering them within the view.

ldecicco-USGS:
nevermind on grid, I like being able to use a side if wanted.

PR #270, opened 2015-10-09
Axis fixes

ldecicco-USGS:


R/access-gsplot.R
```diff
@@ -72,4 +72,28 @@ logged.gsplot <- function(object, side=NULL){
...
+}
+
+#' Get view information from a gsplot object
+#' 
+#' get the views in gsplot object
+#' 
+#' @param object a gsplot object
+#' @return data frame with one row per view. Each view has an x side, y side, the log command, and the view index.
+#' @export
+getViewInfo <- function(object){
```
ldecicco-USGS: @jread-usgs check out this function and tell me if I re-invented some other function

jordansread: no, we don't have something like this. `views(gsplot)` just returns the views (removes the other top level things like par and legend).

ldecicco-USGS: cool. It's kind of handy, might be useful for more than just axis.

jordansread: camelCase..?

ldecicco-USGS: Sorry, hard habit to break. I'll quick change it. We are doing _ for functions and . for arguments, right?

ldecicco-USGS:
OK, the pressure to not reduce the coverage is great. Anyway, I'm done. If someone wants to poke around before merging, cool. Otherwise, you can play with the `axis` examples to see if it acts the way you think it should act.

README.Rmd
```diff
@@ -54,7 +54,8 @@ gs <- gsplot() %>%
   lines(c(3,4,3), c(2,4,6), legend.name="Lines", ylab="logged y axis", xlab="logged x axis", log='xy') %>%
   callouts(x=8, y=4, lwd=2, angle=45, labels="Not sure about this one") %>%
   title("logged axes") %>%
-  axis(side=c(1,2,3,4), labels=FALSE, n.minor=4)
+  axis(side=c(1,2,3,4), labels=FALSE, n.minor=4) %>%
+  axis(side=c(1,2))
```
jordansread, line 58: looks like `grid` happens on top of axis(side=1) in `unnamed-chunk-2-1.png`, which is why there is a diff for that file. Odd that it doesn't matter for side=3.

README.md
```diff
@@ -51,7 +51,8 @@ gs <- gsplot() %>%
   lines(c(3,4,3), c(2,4,6), legend.name="Lines", ylab="logged y axis", xlab="logged x axis", log='xy') %>%
   callouts(x=8, y=4, lwd=2, angle=45, labels="Not sure about this one") %>%
   title("logged axes") %>%
-  axis(side=c(1,2,3,4), labels=FALSE, n.minor=4)
+  axis(side=c(1,2,3,4), labels=FALSE, n.minor=4) %>%
+  axis(side=c(1,2))
 gs
```
jordansread, line 56: :+1: on `unnamed-chunk-3-1.png` minor tick thingy

jordansread:
Looks great to me. I'd say :+1: for merge

README.Rmd
ldecicco-USGS, line 58: I think it's actually due to not ordering the views anymore. We use to put the bgCol first, then grid, then the rest. I'm planning to move those to a top level thing like axis and par (see #239 ) .

PR #276, opened 2015-10-15
xaxs='o' test; fixed pt.lwd/bg issue; cex is formal of callouts.default

lindsayplatt:


tests/testthat/test-set_view_lim.R
```diff
@@ -0,0 +1,41 @@
...
+    points(1:4,1:4) %>%
+    lines(1:6,1:6)
+  
+  beforeBuffer <- c(0,4)
+  buffer <- diff(beforeBuffer)*0.04
+  afterBuffer <- c(beforeBuffer[1], beforeBuffer[2]+buffer)
+  
+  expect_equal(ylim(gs), afterBuffer)
+  expect_equal(par(gs)$yaxs, "i")
+})
```
lindsayplatt, line 41: @ldecicco-USGS @jread-usgs 

Current workflow for yaxs='o' is to include it in the call for points/lines etc. It will buffer the limit of your data in that call, but if you add data after that axis call, it will not be included in the buffer. Should we change the workflow so that it goes in `gsplot()` and therefore applies to all data (see third test for example)?

jordansread, line 41: I think the `'o'` arg belongs at the `view` level, not the gplot level or the points level. In the same way that `ylim` in `points` ends up in `window` for the view, so should `'o'` in my opinion.

lindsayplatt, line 41: Where would you specify it if it doesn't go in the gsplot level? Or would the workflow still be `gsplot(ylim=c(0,NA), yaxs="o")` but it would end up in par?

lindsayplatt, line 41: I think I need a quick chat about this before I can be confident on what I've got - I'm getting lost somewhere between `print` and `calc_views`. I'm not following the new workflow well

lindsayplatt:
@ldecicco-USGS @jread-usgs can we go ahead and merge this? repgen needs some of the first changes, and the only thing that is making this fail is the test for xaxs/yaxs='o'.

ldecicco-USGS:
Sure, more motivation to fix it quick, right?

lindsayplatt:
Yeah, I've spent about an hour and a half or so on it today, but I'm stuck. Could use some fresh eyes - maybe after stand up tomorrow

tests/testthat/test-set_view_lim.R
jordansread, line 41: still in `points` or those types of view things

jordansread, line 41: I am confused by `gsplot(ylim=c(0,NA))`. Isn't that where top-level par args go? ylim is a view arg

lindsayplatt, line 41: I was having trouble carrying over the 'o' option without putting it in par. If it's just in the points call, it only impacts the limits of that call.

PR #284, opened 2015-11-24
barplot (WIP)

lindsayplatt:
- [ ] make tests!!
- [ ] add better examples for documentation
- [ ] work in progress:
- how will legend work since bar plots have their own way `legend.text` and `args.legend`
- currently treating bars/groups of bars as values on x-axis
- using `names.arg` puts names on top of x-axis values
- creates an empty plot and then this (because `add=FALSE`, so `plot.new()` is called) BUT using `add=TRUE` is funky
- lower ylim is not lined up with x-axis

Sort of working:

``` r
fake <- runif(10, 0, 25)
fake_mat <- matrix(fake, nrow=2, byrow=TRUE)

gs_mat <- gsplot() %>% 
    barplot(fake_mat, beside=TRUE, col=c("blue", "green"), 
            density=5, space=c(0.1,0.5), 
            names.arg=c("one", "two", "three", "four", "five"))
gs_mat
```

image


jordansread:
`update.packages()`?

jordansread:
https://github.com/USGS-R/gsplot/blob/master/R/calc_views.R#L123 looks like we need `height` for y. Not sure how we handle width here.

lindsayplatt:
@jread-usgs is that for fixing something?

jordansread:
This might be something, similar to curve, that you will need to break down into its components. Which I think are just rectangle elements. Have you looked into that approach?

lindsayplatt:
We've decided to not implement bar graphs at this time -- users can create them using rectangles

Issue #262, opened 2015-10-06
rename view as side.x.y


jordansread:
came up in discussions/implementation for #246

jordansread:
This makes a lot of things easier I think

jordansread:
This is a big overhaul, but worth it and should probably happen soon.

jordansread:
while we are at it, I wonder if the rendering of each individual side/axis should happen outside of the view. 

For example, this would be a mock-up for the structure of the `gsplot` object:

``` yml
gsplot
   - view.1.2
      - window(...)
      - points(...)
      - lines(....)
   - view.1.4
      - window(...)
      - points(...)
      - points(...)
   - par(...)
   - legend(...)
   - side.1(...)
   - side.2(...)
   - side.4(...)
```

jordansread:
the latter would probably be a separate issue, but I figured I would bring it up here for discussion.

jordansread:
@jiwalker-usgs @ldecicco-USGS did you have thoughts on this one?

jordansread:
we did this

PR #289, opened 2015-12-05
Rename view

jordansread:
For #262 

Comments welcome


R/calc_views.R
```diff
@@ -25,11 +25,12 @@ calc_views <- function(gsplot){
 }
 
 views <- function(gsplot){
-  gsplot[names(gsplot) %in% 'view']
+  gsplot[grep('view.', names(gsplot))]
 }
 
 non_views <- function(gsplot){
-  gsplot[!names(gsplot) %in% 'view']
+  gsplot[grep('view.', names(gsplot))] <- NULL
```
jiwalker-usgs: I'd feel more comfortable here if the argument was copied to an intermediate variable.  Manipulating an argument feels dirty.

R/calc_views.R
```diff
@@ -50,7 +51,9 @@ group_views <- function(gsplot){
       views[[v.i]] <- append(views[[v.i]], to_draw)
       views[[v.i]][['window']][['par']] <- append_replace(views[[v.i]][['window']][['par']], tail.gs[['gs.config']][['par']])
     } else{
-      views <- append(views, list(view = append(to_draw, list(window=list(side=add_sides,par=tail.gs[['gs.config']][['par']])))))
+      new.view <- list(append(to_draw, list(window=list(side=add_sides,par=tail.gs[['gs.config']][['par']])))) %>%
```
jiwalker-usgs, line 55: does this get called in a way that someone doing `side=c(1,2,3,4)` would result in 4 views?  Just wrapping my head around the logic.

R/calc_views.R
```diff
@@ -88,7 +91,7 @@ which_reals <- function(values, na.value){
   
 }
 set_view_window <- function(views, param, na.value=NA, remove=TRUE, ignore=NULL){
-  view_i <- which(names(views) %in% "view")
+  view_i <- grep('view.', names(views))
```
jiwalker-usgs, line 95: 'view.' seems like a magic number that is used in a lot of places, is it possible to make this a constant in some way (not sure if the idea of constants works the same in R packages as in other languages).

R/callouts.R
```diff
@@ -38,10 +38,11 @@ callouts.gsplot <- function(object, ..., side=c(1,2)){
 #' @param labels text to be added to callout
 #' @param length relative (percentage of window width and height) distance for callout
 #' @param angle callout line angle
+#' @param cex passed to text for font size formatting
 #' 
 #' @rdname callouts
 #' @export
-callouts.default <- function(x, y=NULL, labels=NA, length=0.1, angle='auto', cex=1, ...){
+callouts.default <- function(x, y=NULL, labels=NA, length=0.1, angle='auto', cex=par()$cex, ...){
```
jiwalker-usgs, line 45: is this the `par()` at load time, or the `par()` at runtime?  This might be an issue if we change our `par()` scoping.

R/calc_views.R
jordansread, line 55: no, not yet. Right now the side=c(x,y) is a two-side thing. I'd need more clarification on what we expect on side=c(1,2,3,4) - which become the default sides for axis labels? both y and both x? or 1 and 2?

R/calc_views.R
jordansread, line 95: I was thinking more that it would stay as-is, but would happen in only one place: `view_indices <- function(gsplot)` clean up to come on that in the future?

R/callouts.R
jordansread, line 45: it is at render time, just like the other `graphics::` functions we use that have similar defaults. Would be good to verify/test that, but that is the target for this.

R/calc_views.R
jordansread: fixing this. Adding a commit in a min.

jordansread:
@ldecicco-USGS can we merge?

ldecicco-USGS:
Love it, yes

PR #297, opened 2015-12-29
side can be specified for error_bar

lindsayplatt:
@ldecicco-USGS @jread-usgs @jiwalker-usgs did we not have error_bar accept a `side` argument on purpose? Or was it just overlooked or changed at some point and never reverted?


jordansread:
Overlooked I think

R/error_bars.R
```diff
@@ -80,7 +80,7 @@ error_bar.gsplot <- function(object, x, y, y.high=0, y.low=0, x.high=0, x.low=0,
     x.error <- x[errorIndex]
```
jordansread, line 80: Do you need to charge the function formals for error_bars?

lindsayplatt, line 80: @jread-usgs what do you mean by that?

lindsayplatt:
Went ahead and merged because we needed that fix for AQCU. We can revisit @jread-usgs comment next week

PR #299, opened 2016-01-06
bgCol was being called by default

jordansread:


jordansread:
review @lindsaycarr ?

R/bgCol.R
```diff
@@ -62,7 +62,7 @@ bgCol.default <- function(col,...){
     y2 <- par("usr")[4]  
   }
   
-  rect(x1,y1,x2,y2,col = col,...)
+  rect(x1,y1,x2,y2,col = col, border = NA, ...)
```
lindsayplatt, line 65: @jread-usgs is this necessary in addition to checking if bg.arg is null?

jordansread, line 65: I guess I wasn't sure why we would want the line even if `bgCol` was called. We want it to do the rect fill, not draw the border too right?

lindsayplatt, line 65: Yeah, it's probably been over plotting the rectangle line on top of the plot frame this whole time

lindsayplatt:
``` r
gs <- gsplot() %>%
    points(1:4, 1:4) %>%
    bgCol(border=NA, col=NULL)
gs

Error in (function (xleft, ybottom, xright, ytop, density = NULL, angle = 45,  : 
  formal argument "border" matched by multiple actual arguments 
```

`border` is coming through as a named argument & also inside `...`

jordansread:
I don't know why border would be a public arg. Should it be used ever?

lindsayplatt:
It shouldn't be used, but if they pass it in, would we want it to fail?

Issue #130, opened 2015-07-20
Clean up config


ldecicco-USGS:
Make this smarter:

https://github.com/USGS-R/gsplot/blob/master/R/config.R#L60

lindsayplatt:
idea for defining session configs:

`config(type, persist=TRUE, ...)`

lindsayplatt:
@ldecicco-USGS is this still relevant based on recent config work?

PR #300, opened 2016-01-14
Adding a persist parameter

jiwalker-usgs:
to set stuff in the config that will persist for the session.  Reloading config from disk will undo any persisted config.  I'm going to try to tackle #130 but thought this would be a good start.


jordansread:
:shipit:

Issue #302, opened 2016-01-21
legend symbols misplaced when more than one fill exists


lindsayplatt:
legend places symbols in the wrong location when more than one rectangle is added

``` r
#two rectangles
gs <- gsplot() %>% 
  rect(xleft = c(1,2), xright = c(2,4),
       ybottom = c(1,1), ytop = c(2,2),
       col = c("red", "green"), legend.name=c("one", "two")) %>%
  points(5:7, 5:7, pch=24, col="black", bg="grey", legend.name="field visits") %>% 
  legend()
gs
```

![image](https://cloud.githubusercontent.com/assets/13220910/12486430/57647f4a-c028-11e5-86dd-d5ac248a66a1.png)

Compared to:

``` r
gs <- gsplot() %>% 
    rect(xleft = 1, xright = 2,
         ybottom = 1, ytop = 2,
         col = "green", legend.name="one") %>%
    points(5:7, 5:7, pch=24, col="black", bg="grey", legend.name="field visits") %>% 
    legend()
gs
```

![image](https://cloud.githubusercontent.com/assets/13220910/12486449/6c2aaf62-c028-11e5-989e-696201bec92c.png)

lindsayplatt:
My suggested solution to this is to pass in each legend item separately to the `set_legend_args` function. So, when `length(legend.name) > 1`, you should separate the parameters into multiple `paramsAll`'s and pass them through the legend args logic. See my PR that does this #305

lindsayplatt:
This issue has been fixed by #305, but is not the cleanest of solutions. Will open a new issue about making it cleaner

PR #305, opened 2016-01-21
edit formals call for error_bar + handling multiple legend.names

lindsayplatt:
fixing #302


R/set_args.R
```diff
@@ -54,6 +54,23 @@ set_legend_args <- function(object, fun.name, ..., legend.name) {
   
   paramsAll <- set_args(fun.name, list(...))
   
+  if(length(legend.name) > 1){
+    paramsAll_df <- as.data.frame(paramsAll, stringsAsFactors = FALSE)
```
lindsayplatt, line 58: I would like some input on a more elegant way to accomplish this - trying to separate the list of params into multiple lists of params. E.g. if `listAll <- list(x=c(1,2), y=c(2,3), col=c("red", "green"), legend.name = c("point 1", "point 2"))` then I would like to be able to send `list1 <- list(x=1, y=2, col="red", legend.name="point 1")` and `list2 <- list(x=2, y=3, col="green", legend.name="point 2")` to `get_legend_args` separately.

lindsayplatt:
@jiwalker-usgs this is what I was talking about last night. Have some feedback?

R/set_args.R
jiwalker-usgs, line 58: Would it work to move this logic into get_legend_args?  It seems like it shouldn't be the responsibility of set_args to know quirks that exist with legends.  I think here we just need to be able to handle whatever get_legend_args hands back.  I'm not sure that is right though.

lindsayplatt, line 58: So it is inside of `set_legend_args`. I created `get_legend_args` as part of this PR to separate the logic that figures out what should be a `legend.args` list item from what is creating the elements inside that list. `set_args` is being called from `set_legend_args` to get the parameters that are being plotted.

lindsayplatt:
@jread-usgs @jiwalker-usgs just committed a draft solution for handling multiple callout angles. I would love it if someone could suggest a vectorized substitute to my for loop on L115-120.

lindsayplatt:
oh and comments on the commits for splitting up the legend.args -- I am blanking on the suggestions you gave Friday @jiwalker-usgs (should've written them down!)

R/callouts.R
```diff
@@ -66,30 +67,58 @@ callouts.default <- function(x, y=NULL, labels=NA, length=0.1, angle='auto', cex
...
+    angle <- x1_y1_angle$angle
   } else {
-    auto.angle <- c(30, 330, 150, 210)
-    x1 <- sapply(auto.angle, function(a) {
-      x + length * xrange * cos(2*pi*(a/360))
-    })
-    y1 <- sapply(auto.angle, function(a) {
-      y + length * yrange * sin(2*pi*(a/360))
-    })
+    stopifnot(length(x)%%length(angle) == 0) #stop if the defined angles aren't a multiple of the points
```
jiwalker-usgs, line 108: Can we print warnings or something to point out where and what the problem is?

R/set_args.R
```diff
@@ -54,6 +54,23 @@ set_legend_args <- function(object, fun.name, ..., legend.name) {
...
+
+    for(p in seq(nrow(paramsAll_df))) {
```
jiwalker-usgs, line 60: I think I was suggesting moving this loop to the get_legend_args and having it handle 1 or many legend.names, but I'm not sure the best approach to legend stuff at this point.

jiwalker-usgs:
I'd say this is good at this point.  We'll probably want to circle back to legend stuff at some point to make it clean and more obvious what is going on, but good from here.

R/callouts.R
lindsayplatt, line 108: Doesn't this automatically print a warning? If it isn't true, it will say "length(x) %% length(angle) is not equal to zero". Or are you saying make it a little more user-friendly?

lindsayplatt:
@jiwalker-usgs thanks for the review - for the sake of repgen, I am going to go ahead and merge this. I will close the issue and open a different one about cleaning it up.

lindsayplatt:
Whoops - didn't mean to close this PR

PR #309, opened 2016-01-30
Config cleanup

jiwalker-usgs:
I wasn't sure how best to test this, so I'm not sure whether the coverage is good enough.  Related to #130


tests/testthat/tests-config.R
```diff
@@ -20,3 +20,9 @@ test_that("persisting to config alters environment", {
   expect_equal(gsplot:::config("points")$col, "blue")
   cleanup()
 })
+
+test_that("formals are correctly retrieved", {
+  expect_equal(length(config("par")), 6)
+  expect_equal(names(config("points")), c("pch", "col"))
+  expect_equal(length(config("arrows")), 0)
+})
```
lindsayplatt, line 28: Should we add a test for something that's not allowed and see what it should return?

jiwalker-usgs, line 28: Good idea, like `config("blah")` and expect an error

lindsayplatt, line 28: precisely :smile:

Issue #310, opened 2016-02-17
n.minor not working as expected


lindsayplatt:
`n.minor` uses the default axis `at <- axTicks(axisParams$side)` to calculate where ticks should be. This `at` should be overwritten by any user specified `at`. 

``` r
#works as expected when user does not specify at
gs <- gsplot() %>% 
  points(x = 1:3, y = 1:3) %>% 
  axis(side=1, n.minor= 1)
gs
```

![image](https://cloud.githubusercontent.com/assets/13220910/13118482/fc041bea-d569-11e5-826f-2127c37ea2b3.png)

``` r
#doesn't work as expected when user specifies at
gs <- gsplot() %>% 
  points(x = 1:3, y = 1:3) %>% 
  axis(side=1, at=1:3, n.minor= 1)
gs
```

![image](https://cloud.githubusercontent.com/assets/13220910/13118469/ec291c16-d569-11e5-8654-27a2499a964d.png)

lindsayplatt:
Fixed with #311

PR #311, opened 2016-02-17
axis: n.minor use auto axis ticks if "at" hasn't been user-specified

lindsayplatt:
Addressing issue #310

### DIFF
--- a/R/axis.R
+++ b/R/axis.R
@@ -96,8 +96,16 @@ draw_axis <- function(gsplot, index.axis) {
     } else {
       axis(axisParams)
       n.minor <- n.minor + 1
-      #Minor axis:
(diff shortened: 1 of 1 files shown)


lindsayplatt:
@ldecicco-USGS @jread-usgs @jiwalker-usgs would anyone like to give this a quick review and then merge?

jiwalker-usgs:
Makes sense to me.  Unrelated to this fix, I think we eventually want to move this to the gsplot object so that ticks and minor ticks are just part of the config.  I might be pushing too far in that direction though.

lindsayplatt:
travis and appveyor never kicked off for this...can you manually kick them off without a new commit?

Issue #290, opened 2015-12-05
Pull single side functions out of view


jordansread:
This is touched in discussion for #262 , but I'd like to make it a formal proposal w/ this issue. 

Here is what I am thinking:
`views` should only contain things that require the geometry of 2 (or more) sides to be defined
`sides` (a new top-level addition to the object) would contain things that are only relevant to a single side, such as `axis` calls, `mtext` calls, etc
`par` and `legend` stay as-is, as top-level. 

So an object like this:

``` r
gs <- gsplot()  %>% 
     points(1:20, 3:22, col="blue", legend.name="Points 1") 
```

would have this:

``` r
names(gs)
[1] "view.1.2" "par"      "legend"      "side.1"      "side.2"
```

These kinds of things: https://github.com/USGS-R/gsplot/blob/master/R/print.R#L83 would then be handed in the `sides` functions

jordansread:
Likewise, 

``` r
gs <- gsplot()  %>% 
     points(1:20, 3:22, col="blue", legend.name="Points 1") %>% 
     points(20:1, 3:22, side=c(3,2), col="blue", legend.name="Points 1")
names(gs)
[1] "view.1.2" "view.3.2" "par"      "legend"        "side.1"      "side.2"      "side.3"
```

lindsayplatt:
I like this idea, but if you aren't explicitly making a call to a function that only impacts one side (axis, mtext, etc), would the top-level single sides need to be there?

jordansread:
I think they would be there no matter what. This way we can store some extra info in there, like when a user specifies xlim for a side, there can be a slot for it. Would be `NULL` by default

jordansread:
plus, if you don't call axis, the `side.1` is where the default axis call would live.

jiwalker-usgs:
I'd also like to add to this that we try to not have duplicate data as much as possible.  So rather than view.1.2 having `xlim` and `ylim`, side.1 has `lim` and side.2 has `lim` and view.1.2 knows to go there for `xlim` and `ylim`.

jordansread:
yes, exactly. Fixes the overplotting issue too

lindsayplatt:
That axis overplotting thing is really frustrating to look at. My plots look _almost_ perfect

jordansread:
so close

jordansread:
fixed w #313

PR #313, opened 2016-04-10
Iso sides

jordansread:
WIP for #290 

This is a partial for that, because it 
- [x] ~~doesn't include putting logs in the sides (they are in `view.x.y.$window` still)~~ 
- [x] ~~breaks a number of tests right now (~8 failures)~~
- [x] ~~changes the `logged` functions to return `sides.n` instead of `sides.x.y` which is probably an API-breaker for some if they were using that~~ this is known, but would like someone to weigh in on it
- [ ] doesn't use the `yaxs` and `xaxs` args. The logic for those is still in the PR, but is commented out. I wasn't sure what was going on before
- [x] ~~doesn't include documentation on the new changes/function~~
- [x] ~~no tests for the new stuff~~
- [x] ~~sides aren't used in the rending of axis, so we still have overplotting~~.
- [x] ~~breaks `draw_custom_grid` which is based on windows, not sides~~
- [x] ~~grid is failing when you don't have a side defined for it (e.g., have 1,2 and want axis side 3 w/ grid)~~
- [x] ~~labels are still in window and not sides~~

But, this does get us most of the way to #290 issue, and also I think addresses some speed issues I ran into when creating very large lists. **now**, the `set_side_lim()` (was `set_view_lim()`) only extracts args from the most recent addition to the list, instead of all (which is what was happening before). It also short-circuits the limit extraction when the user limits are fully defined (min and max). I think both of these additions will make it quite a bit faster, but I haven't tested. 

Also, I noticed that the window$log function isn't really working as expected, and I don't think it has anything to do with this PR. For example, when you have view.1.2 and view.3.2 and log y for one of those, it doesn't also log y for the other view. This causes one of the test failures due to the way I am using `logged` now. 

feedback welcome. I will keep this open while I chip away at some of the remaining pieces.


R/calc_views.R
```diff
@@ -123,73 +105,6 @@ set_view_lab <- function(views){
...
-      
-      data.lim <- range(data.var[is.finite(data.var)])
-      usr.lim <- views[[n.i]][['window']][[lim.name]][1:2]
-      views[[n.i]][['window']][[lim.name]] <- data.lim
-      views[[n.i]][['window']][[lim.name]][!is.na(usr.lim)] <- usr.lim[!is.na(usr.lim)]
-    
-      
-      usr.axs <- axs[[axs.name]][[n.i]]
-      
-      if (any(!is.na(usr.axs)) && any(usr.axs == 'o')) {
```
jordansread, line 171: This is part that I didn't really get. @lindsaycarr can you help me think of how to pull this into the `sides` model?

R/access-gsplot.R
```diff
@@ -12,14 +12,28 @@ xlim <- function(object, side) UseMethod("xlim")
 
 #' @export
 xlim.gsplot <- function(object, side=NULL){
+  side.lim(object, side, 'x')
+}
 
+side.lim <- function(object, side, axis = c('x','y')){
```
lindsayplatt: move down so that x and y are next to each other

R/access-gsplot.R
```diff
@@ -12,14 +12,28 @@ xlim <- function(object, side) UseMethod("xlim")
...
+  else {
+    sides <- as.side(names(sides(object)))
+    if (axis == 'y')
+      use.sides <- sides %% 2 == 0
+    else 
+      use.sides <- !sides %% 2 == 0
+    side.names <- as.side_name(sides[use.sides])
+  }
+  
+  lims <- lapply(side.names, function(x) object[[x]]$lim) %>%
```
lindsayplatt: don't use %>% inside of the function

R/calc_views.R
```diff
@@ -23,42 +19,35 @@ calc_views <- function(gsplot){
...
-  views <- views(gsplot) # existing
-  non.views <- non_views(gsplot)
   add_sides <- set_sides(tail.gs[['gs.config']][['side']])
-    
+  non.views <- non_views(gsplot, include.sides = FALSE)
+  vew.n.sde <- gsplot[c(which_views(gsplot), which_sides(gsplot))]
   if (!is.null(add_sides)){
+    vew.n.sde <- append_sides(vew.n.sde, add_sides)
     to_draw <- setNames(list(c(tail.gs[['arguments']], legend.name=tail.gs[['gs.config']][['legend.name']])), tail.nm)
+    view.name <- sprintf('view.%s.%s',add_sides[1],add_sides[2])
```
lindsayplatt: use `as.view_name` fucntion

R/callouts.R
```diff
@@ -45,8 +45,7 @@ callouts.gsplot <- function(object, ..., side=c(1,2)){
 callouts.default <- function(x, y=NULL, labels=NA, length=0.1, angle='auto', cex=par()$cex, ...){
   
   if (is.null(y)) {
-    warning("y=NULL not currently supported in callouts.default")
-    return()
+    stop("y=NULL not currently supported in callouts.default", call. = FALSE)
```
lindsayplatt: revert please

PR #330, opened 2016-04-18
WIP for add_to_view

jordansread:
don't merge plz


R/function_defaults.R
```diff
@@ -0,0 +1,18 @@
+
+function_defaults <- function(fun.name, out=c('package','def.funs')){
+  funs <- list('points'=
```
jiwalker-usgs, line 3: There is a list similar to this in config().  We may want to share this info with that  to reduce duplication.

jordansread, line 3: Good thought. Will look there

jordansread, line 3: Yeah, looks like we are doing different things for formals between these functions, and that should be consistent. For these functions, I take formals from both `plot.xy` and `lines.default`. In config it just uses one of those

R/add_to_view.R
```diff
@@ -0,0 +1,58 @@
...
+#'          points(x=1:5, y=1:5, xlim=c(0,10), ylim=c(0,10), 
+#'                callouts(labels=c(rep(NA, 4), "oh")), 
+#'                error_bar(y.high=1))
+#' gs
+#' 
+#' gsplot:::call_arguments('points', x=2:6, y=2:6, ylim=c(-1, 11))
+#' gsplot:::call_arguments('points', x=1:5, y=1:5, xlim=c(0,10), ylim=c(0,10), 
+#'                callouts(labels=c(rep(NA, 4), "oh")))
+#' @keywords internal
+call_arguments <- function(fun.name, ...){
```
lindsayplatt, line 17: @jread-usgs would this return `legend.args` if it was specified in the call?

R/add_to_view.R
```diff
@@ -0,0 +1,58 @@
...
+#' 
+#' @param fun.name
+#' @param .dots lazy_dots
+#' @keywords internal
+normal_arguments <- function(fun.name, .dots){
+
+  fun.defaults <- function_defaults(fun.name)
+  
+  if (!is.null(.dots))
+    arguments = set_args(fun.name, lazy_eval(.dots), package=fun.defaults$package)
```
lindsayplatt, line 36: @jread-usgs I am getting an error at this line when I run the example in the gsplotIntro.Rmd. The error is only for abline, works fine when I comment it out. This only happens when abline has `legend.name` defined. 

``` r
demoPlot <- gsplot() %>%
  points(y=c(3,1,2), x=1:3, xlim=c(0,NA),ylim=c(0,NA),
         col="blue", pch=18, legend.name="Points", xlab="Index") %>%
  lines(c(3,4,3), c(2,4,6), legend.name="Lines", ylab="Data") %>%
  abline(b=1, a=0, legend.name="1:1")

Error in asNamespace(ns) : not a namespace
```

lindsayplatt, line 36: ugh, this is also happening with `rect`

R/add_to_view.R
jordansread, line 17: you mean `legend.name`?

R/add_to_view.R
jordansread, line 36: hmmm...I thought I ran the vignettes w/ v0.6.0

R/add_to_view.R
lindsayplatt, line 17: yes I did, but you answered that question at our review yesterday :)

R/add_to_view.R
lindsayplatt, line 36: yeah, pretty sure it's from the changes I have made but I am not understanding how the error would occur here

Issue #236, opened 2015-09-10
careful to pick up defaults when we bypass a function


jordansread:
Skipping defaults w/ values (e.g., for plot.xy lty = par("lty")) doesn't matter because `graphics::function` is called ultimately, and those defaults get in there. 

but...bypassing functions like `curve` to create a line call doesn't actually grab the defaults of `curve` (e.g., `n=101`).

We should work those in whenever we bypass.

lindsayplatt:
@jread-usgs can you give an example? Not sure how to test if this is fixed already or not.

jordansread:
yes, our implementation of `curve` was an example of this. But you fixed it: https://github.com/USGS-R/gsplot/commit/7069295685c2825b45af9baeb53731cdc7e7c540

lindsayplatt:
Ah I see, this is case by case. So something we need to keep in mind if we add something that bypasses. Going to leave this open so that we can verify that we don't have anymore occurrences, but not going to add to milestone 0.8.0

PR #335, opened 2016-04-19
modify_legend (WIP)

lindsayplatt:


lindsayplatt:
- modify_legend should create one legend.args list element (not append multiple to the gsplot object)
- draw_legend currently mushes all of the legend.args together, but this should be handled in modify_legend
- draw_legend should only be calling the graphics `legend()`

lindsaycarr:
currently breaks everything that is not points because `modify_legend` takes diff arguments than `set_legend_args` did

lindsayplatt:
Currently breaks everything that is not points because modify_legend takes diff arguments than set_legend_args did. 

To do tomorrow so that it is no longer broken:
- [x] get the rest of set_legend_args functionality into modify_legend (multiple legend args in one function call, etc)
- [x] figure out how/when to do overall legend stuff (when are configs like location and border set)
- [ ] change set_legend_args to modify_legend in all function calls
- [ ] update tests

lindsayplatt:
failing on abline and rect calls: fails at `modify_legend >> call_arguments >> normal_arguments >> set_args >> function_args >> getFromNamespace`

lindsayplatt:
The following test passes when I replace NA with NA_integer_, but that is not a new test. I'm wondering why it did not fail before. New package update maybe? 

``` r
gs <- gsplot() %>%  
    points(x=1, y=2, legend.name="Points 1", pch=1, col="blue") %>% 
    points(x=3, y=4, legend.name="Points 2", pch=5, col="red") %>%  
    error_bar(3,4, y.high=0.2, y.low=0.3, legend.name="Errors", text.font=2) %>% 
    rect(2,3,2.5,4, col="blue", border="orange", lwd=3, density=15, legend.name="rectangle")  %>% 
    legend(location="bottomright", bg="lightgrey")

    expect_equal(gs$legend$legend.args$lwd[4], NA)
```

This one is failing somewhere in call_arguments >> normal_arguments >> set_args

``` r
gs <- gsplot() %>%
     curve(sin(x), from=-2*pi, to=2*pi, legend.name="sin(x)") %>%
     legend()

Error in eval(expr, envir, enclos) : could not find function "*"
```

jordansread:
this is really bizarre - the environment for the lazy_eval is null when you do something else w/ the dots 

``` r
eg <- function(...){
     print(names(list(...)))
     lazyeval::lazy_eval(lazyeval::lazy_dots(...))
}

eg(from=2 * pi)
[1] "from"
 Error in eval(expr, envir, enclos) : could not find function "*" 

#vs comment out the `print`:
eg <- function(...){
     #print(names(list(...)))
     lazyeval::lazy_eval(lazyeval::lazy_dots(...))
}
eg(from=2 * pi)
$from
[1] 6.283185
```

not what I would expect to happen. 

This is a majorly simplified example of what is happening w/ that test in your PR.

jordansread:
The lazy eval thing is something I don't understand. But the lazy dots things in curve aren't actually needed so that odd behavior can be avoided. 

I think `curve.gsplot` can be re-written as:

``` r
curve.gsplot <- function(object, expr, from=0, to=1, n=101, ..., legend.name=NULL, side=c(1,2)){

  expr <- lazy(expr)

  x <- seq(from, to, length.out=n)
  y <- lazy_eval(expr, data.frame(x=x))
  object <- lines(object, x=x, y=y, ..., legend.name=legend.name, side=side)
  return(object)
}
```

That fixes the test issue you are running into and deals with the minor issues pointed out in #236

R/modify_legend.R
```diff
@@ -0,0 +1,191 @@
...
+#'          points(x=1:2, y=1:2, col = c("red", "blue"), 
+#'                 legend.name = c('points 1', 'points 2')) %>% 
+#'          legend()
+#' gs
+#' 
+#' @keywords internal
+add_to_legend <- function(object, fun.name, legend.name, ...){
+  
+    if(is.null(legend.name)) {
+      return(object)
```
jordansread, line 24: the case where you don't have a legend.name, but you do call legend results in an error:

``` r
gs <- gsplot() %>%
   points(1,2) %>%
   legend()
gs
 Error in (function (x, y = NULL, legend, fill = NULL, col = par("col"),  : 
  unused argument (draw = TRUE)
```

lindsayplatt:
@jread-usgs added a fix for calling legend() without any legend.names. For curve, should we do something different for the n default? Like keep `increment <- (arguments$to-arguments$from)/10000` this?

lindsayplatt:
Oh, just noticed that `n=101` is the actual default. I leave that be. 

When n is specified is causes an error:

``` r
gs <- gsplot() %>%
    curve(sin(x), from=-2*pi, to=2*pi, n=10, legend.name="sin(x)") %>%
    legend()
gs

Error in function_args(package, name, object, ...) : object 'x' not found 
```

jordansread:
Hmm - I can't tell where you modified the curve.R file. Is that in here?

lindsayplatt:
@jread-usgs just pushed from local

jordansread:
@lindsaycarr all good to merge?

lindsayplatt:
Yay! Just saw this :)

jordansread:
We'll need to be careful with capturing the dots in proper order to make sure we don't run into that null env issue: https://github.com/hadley/lazyeval/issues/61

R/modify_legend.R
```diff
@@ -0,0 +1,191 @@
...
+#' gs
+#' 
+#' gs <- gsplot() %>% 
+#'          points(x=1:2, y=1:2, col = c("red", "blue"), 
+#'                 legend.name = c('points 1', 'points 2')) %>% 
+#'          legend()
+#' gs
+#' 
+#' @keywords internal
+add_to_legend <- function(object, fun.name, legend.name, ...){
```
jordansread, line 21: @lindsaycarr can you help me figure out how to plug this in? 

For example, should a function like `points` call both `modify_legend` and `add_to_lengend`?

PR #336, opened 2016-04-19
Adding indices to `sides()`

jiwalker-usgs:


jiwalker-usgs:
@jread-usgs does this make any sense to you?

R/utils-side.R
```diff
@@ -85,8 +85,14 @@ which_sides <- function(gsplot){
 #' @param gsplot a gsplot object
 #' @return a sides list
 #' @keywords internal
-sides <- function(gsplot){
-  gsplot[which_sides(gsplot)]
+sides <- function(gsplot, by.index=NA){
+  mySides <- gsplot[which_sides(gsplot)]
+  if (!all(is.na(by.index))) {
+    split <- strsplit(names(mySides), split = "side.", fixed = TRUE)
```
jordansread: do we not have a `as.side()` method that works on the names of the sides? If so, we should use it here, otherwise will be a pain if we change out side naming convention

R/utils-side.R
```diff
@@ -85,8 +85,14 @@ which_sides <- function(gsplot){
...
+    indices <- sapply(split, function(x) x[2])
+    mySides <- mySides[indices %in% by.index]
```
jordansread, line 92: otherwise, yeah, makes sense

jiwalker-usgs:
fixed that

PR #342, opened 2016-05-02
Overhaul for our modify work

jordansread:
this is a WIP, but I think I need some help from @jiwalker-usgs to converge on how we are stitching this together


R/add_to_view.R
```diff
@@ -46,36 +56,46 @@ add_to_view <- function(object, call.args, side){
...
 #' 
-#' gsplot:::call_arguments('points', x=2:6, y=2:6, ylim=c(-1, 11))
-#' gsplot:::call_arguments('points', x=1:5, y=1:5, xlim=c(0,10), ylim=c(0,10), 
-#'                callouts(labels=c(rep(NA, 4), "oh")))
+#' gsplot:::filter_arguments('points', x=2:6, y=2:6, ylim=c(-1, 11))$call.args
+#' gsplot:::filter_arguments('points', x=1:5, y=1:5, xlim=c(0,10), ylim=c(0,10), 
+#'                callouts(labels=c(rep(NA, 4), "oh")))$extracted.args
 #' @keywords internal
-call_arguments <- function(fun.name, ...){
+filter_arguments <- function(fun.name, ...){
```
jordansread, line 63: since some functions use only call.args or maybe ignore embedded args, I would also like to think about adding an additional arg as `out=c("call.args","option.args","extracted.args")` by default. Thoughts? So `bgCol` would use `out=c("call.args")`

jordansread:
@jiwalker-usgs do you have time to review? This is still a WIP, but could use some feedback

R/add_to_view.R
```diff
@@ -1,4 +1,13 @@
 
+apply_extracted_args <- function(object, extracted.args, side=c(1,2)){
```
jiwalker-usgs, line 2: side seems to not be used in this function, not sure if it is missing.

R/add_to_view.R
```diff
@@ -84,14 +104,14 @@ normal_arguments <- function(fun.name, .dots){
 #' @param embedded.dots expressions to be evaluated within \code{parent.args} data
 #' @param parent.args data that should be accessible when evaluating \code{embedded.dots}
 #' @keywords internal
-embedded_arguments <- function(fun.name, embedded.dots, parent.args){
+nonstandard_eval_arguments <- function(fun.name, embedded.dots, parent.args){
```
jiwalker-usgs, line 107: I like this name change, embedded was confusing

R/axis.R
```diff
@@ -67,23 +67,22 @@ axis.gsplot <- function(object, ..., n.minor=0, tcl.minor=0.15, reverse=NULL) {
...
+                       gs.config=list(n.minor=n.minor, tcl.minor=tcl.minor, 
+                                      reverse=reverse))
+    to.gsplot <- list(arguments1)
+    names(to.gsplot) <- fun.name
     
     object <- append(object, to.gsplot)
   }
-  
-  return(gsplot(object))
+  class(object) <- 'gsplot'
```
jiwalker-usgs, line 84: This feels weird to me, the old way seems more logical.

R/calc_views.R
```diff
@@ -1,206 +0,0 @@
-#' calc_view
```
jiwalker-usgs, line 1: Nice to see this removed :)

R/gather_function_info.R
```diff
@@ -0,0 +1,22 @@
+
+
```
jiwalker-usgs: This probably needs some description as it is used pretty heavily.

R/order_views.R
```diff
@@ -1,27 +0,0 @@
-set_view_order <- function(views, order){
```
jiwalker-usgs, line 1: do we have anything to help with ordering now?

R/print.R
```diff
@@ -94,13 +100,12 @@ print.gsplot <- function(x, ...){
...
 print.view <- function(x, ...){
-  plots <- remove_field(x, param = c('window','grid'))
-  window <- x[['window']]
+  plots <- remove_field(x, param = c('par','grid','window'))
   
-  if(window$frame.plot){
-    box()
-  } 
-  par(window[['par']])
+  # if(window$frame.plot){
```
jiwalker-usgs, line 106: is this something we are losing? do we need it?

R/utils-list.R
```diff
@@ -0,0 +1,70 @@
...
+      } else
+        out <- append(out, NA)
+      
+    }
+    
+  }
+  return(out)
+}
+
+strip_pts2 <- function(data, param) {
```
jiwalker-usgs, line 54: Do we have any good tests (or idea of what they should be) for what strip_pts should be?  I tried to make replace it with this function, but wasn't sure what I might be breaking.

R/axis.R
jordansread, line 84: yes, good catch. I was doing that because gsplot(object) was sending it through calc_views (not happening anymore). will revise this.

R/print.R
jordansread, line 106: Yes, and I wanted to bring up where we want to capture/store this. It is the boolean for box=TRUE but it isn't a par argument, it is a plot.window argument. It probably should be within view, but I punted on it for now

R/order_views.R
jordansread, line 1: no we don't, other than separating sides and views vs the globals. We might want to re-think how ordering should be applied and enforced. Right now it is first come first plotted except for axis and legend (first and last).

Issue #344, opened 2016-05-11
bugginess from major overhaul


lindsayplatt:
#342 made big changes, but passed tests

Here is a collection of things that no longer work as expected (and therefore need tests added). Details in the comments below.
- [ ] ~~no box around the plots~~ we know this is an issue, blocked by #350 
- [x] `las` defined in default.yaml not being used
- [x] oddly thick legend box
- [ ] ~~axis lines not extended to the actual limits~~ this is base behavior
- [ ] ~~legend not correctly spaced away from x axis label~~ this was an issue in previous versions
- [ ] ~~custom legend ability is gone~~ pulled out into #353

lindsayplatt:
- [ ] ~~no box around the plots~~ we know this is an issue, blocked by #350 
- [x] `las` defined in default.yaml not being used
- [x] oddly thick legend box

Example plot that shows these issues:

``` r
MaumeeDV <- MaumeeDV

demoPlot <- gsplot() %>%
  points(y=c(3,1,2), x=1:3, xlim=c(0,NA),ylim=c(0,NA),
         col="blue", pch=18, legend.name="Points", xlab="Index") %>%
  lines(c(3,4,3), c(2,4,6), legend.name="Lines", ylab="Data") %>%
  abline(b=1, a=0, legend.name="1:1") %>%
  legend(location="topleft",title="Awesome!") %>%
  grid() %>%
  error_bar(x=1:3, y=c(3,1,2), y.high=c(0.5,0.25,1), y.low=0.1) %>%
  error_bar(x=1:3, y=c(3,1,2), x.low=.2, x.high=.2, col="red",lwd=3) %>%
  callouts(x=1, y=2.8, lwd=2, angle=250, labels="Weird data") %>%
  title("Graphing Fun")
demoPlot
```

![image](https://cloud.githubusercontent.com/assets/13220910/15189422/38aa3d98-1770-11e6-9db9-afa4a7a376dd.png)

lindsayplatt:
- [ ] ~~axis lines not extended to the actual limits~~ that's actually normal (when a box is not drawn around, it looks like that, even in base: `plot(dates[[site]], Wtemp[[site]], col="orangered", type='l', frame.plot=F)`
- [ ] ~~legend not correctly spaced away from x axis label~~ this was an issue in previous versions

Example plot that shows these issues:

``` r
sites <- unique(MaumeeDV$site_no)
[...]

PR #345, opened 2016-05-11
updates after review

jordansread:
per #344 

@lindsaycarr the `box()` framing of the plot still isn't active here, and I am still wondering where the best place to store that is. 

Also "axis lines not extended to the actual limits" that is part of the box being missing, so I didn't mess with that. The other issues were related to par slipping out of scope and affecting legend defaults. 

Can you take a look?

### DIFF
--- a/R/print.R
+++ b/R/print.R
@@ -53,7 +53,7 @@ print.gsplot <- function(x, ...){
 
   for (side.name in side.names){
     side <- as.side(side.name)
-    par(x[[side.name]]$par)
+    old.par <- par(x[[side.name]]$par)
     set_frame(views, side)
     if(!side %in% defined.sides){ 
       if(x[[side.name]][['axes']]){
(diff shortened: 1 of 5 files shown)


R/print.R
```diff
@@ -53,19 +53,20 @@ print.gsplot <- function(x, ...){
 
   for (side.name in side.names){
     side <- as.side(side.name)
-    par(x[[side.name]]$par)
+    old.par <- par(x[[side.name]]$par)
     set_frame(views, side)
     if(!side %in% defined.sides){ 
       if(x[[side.name]][['axes']]){
-        Axis(side=side,x=lim(views, side))
+        do.call('Axis', append(list(side=side,x=lim(views, side)), config('axis')))
```
lindsayplatt, line 60: what does having a do.call here do?? Isn't it just applied to one list element anyways?

jordansread, line 60: you get this:

``` r
Axis(side=2,x=c(1,0), config('axis'))
Error in axis(side = side, at = at, labels = labels, ...) : 
  (list) object cannot be coerced to type 'double'
```

jordansread, line 60: can be done by this: 

``` r
Axis(side=2,x=c(1,0),NULL,config('axis'))
```

since usage of `Axis` is `Axis(x = NULL, at = NULL, ..., side, labels = NULL)`
but...I avoided that because `at` could potentially be in config('axis') although that would be unlikely

jordansread:
@lindsaycarr are you 👍 on this? If so, can you merge?

lindsayplatt:
Oops, I always forget that last important step!

Issue #321, opened 2016-04-17
reverse doesn't work anymore


jordansread:
``` r
gs <- gsplot() %>%
   points(1:10, 1:10) %>%
   axis(1, at = seq(0,10,by=0.1),labels=FALSE, tcl=0.15) %>%
   axis(2, reverse=TRUE)
gs
```

![image](https://cloud.githubusercontent.com/assets/2349007/14586929/99d05e92-046d-11e6-8d04-fafc8ea13b5e.png)

no test for this as far as I could tell

jordansread:
Should probably happen within `lim` function, which would use `rev` if the axes is reverse, but this would be way easier and cleaner if `axis` were contained within their appropriate side, re #320

lindsayplatt:
This seems to be fixed in 0.6.1, but still no test

PR #347, opened 2016-05-12
Moving axis calls to sides

jordansread:


jordansread:
I think you'll like those image diffs @ldecicco-USGS (`las`)

ldecicco-USGS:
:heart_eyes:

R/access-gsplot.R
```diff
@@ -113,8 +113,14 @@ lim <- function(object, side=NULL, axis = NULL, set.undefined=TRUE, if.null=c(0,
...
-  lims <- lapply(side.names, function(x) object[[x]]$lim) %>% 
-    setNames(side.names)
+  lims <- lapply(side.names, function(x) {
+    lim <- object[[x]]$lim
+    if (object[[x]]$reverse){
+      lim <- rev(lim)
+    }
+    return(lim)
+  })
+  names(lims) <- side.names
```
jordansread, line 123: This should cover #321 but maybe could be done in a better way ...

R/axis.R
```diff
@@ -71,79 +71,78 @@ axis.gsplot <- function(object, ..., n.minor=0, tcl.minor=0.15, reverse=NULL) {
...
-    arguments1 <- list(arguments = append(list(side=i), user_args[[fun.name]]), 
-                       gs.config=list(n.minor=n.minor, tcl.minor=tcl.minor, 
-                                      reverse=reverse))
-    to.gsplot <- list(arguments1)
-    names(to.gsplot) <- fun.name
+  for(side in sides){
+    # append the side and give it defaults if it doesn't exist
+    object <- modify_side(object, args = list(), side=side) 
     
-    object <- append(object, to.gsplot)
```
jordansread, line 82: we don't have `add_to_side`, which is really what is happening here, but it isn't just adding it as a new call, it is replacing the default `axis` call that is placed in when you create a new side. I wasn't quite sure how we should do that.

R/print.R
```diff
@@ -88,8 +83,8 @@ print.gsplot <- function(x, ...){
     par(new=TRUE)
   }
   
-  i.axis.noview <- i.axis[which(!defined.sides %in% c(view.info$x, view.info$y))]
-  draw_axis(views, index.axis=i.axis.noview)
+  #i.axis.noview <- i.axis[which(!defined.sides %in% c(view.info$x, view.info$y))]
```
jordansread, line 86: I don't really know why this was here or what purpose it served.

R/utils-side.R
```diff
@@ -222,7 +222,10 @@ add_new_side <- function(object, side.name){
   stopifnot(length(side.name) == 1)
   if (side.name %in% side_names(object))
     stop(side.name, ' already exists, cannot add it.', call. = FALSE)
-  side.template <- list(list(lim = c(NA, NA), log=FALSE, label="", axes = TRUE, reverse = FALSE, usr.lim=c(FALSE, FALSE)))
+  side.template <- list(list(
```
jordansread, line 225: The side template now has a default axis call built in automatically that is overridden if a user calls `axis`

jordansread:
@ldecicco-USGS @aappling-usgs can you build box plot fig with this version? `0.6.1`

aappling-usgs:
sounds good. can i merge this?

jordansread:
I think so 👍

PR #348, opened 2016-05-13
barplot work in progress

lindsayplatt:
Lots issues because of barplot not really having an x-axis and its own legend stuff. You can't add on addition gsplot features because of the x-axis thing. A very simple example works though:

``` r
gs <- gsplot() %>%
    barplot(height = cbind(x = c(465, 91) / 465 * 100,
                           y = c(840, 200) / 840 * 100,
                           z = c(37, 17) / 37 * 100),
            beside = FALSE,
            width = c(465, 840, 37),
            col = c(1, 2))    %>% 
    title("Barplot in gsplot")
gs
```

image


lindsayplatt:
Will come back to at a later date

PR #355, opened 2016-06-29
code for several issues

jordansread:


man/bgCol.default.Rd
```diff
@@ -4,14 +4,12 @@
 \alias{bgCol.default}
 \title{create a background color}
 \usage{
-bgCol.default(col, ...)
```
jordansread, line 7: took this out because `bgCol` should only have 1 arg, right? unless we are supporting shading. If we are (like w/ `density`) it should be an explicit arg, not `...`

R/print.R
```diff
@@ -73,13 +73,15 @@ print.gsplot <- function(x, ...){
     if(any(names(views[[view.name]]) %in% 'grid')){
       draw_custom_grid(views,view.name)
     }
-    print.view(views[[view.name]])
-        # -- initial view --
+    # -- initial view --
     if(view.name == view_names(views)[1]){
       if (!is.null(bg.arg))
```
jordansread: do we need this test?

R/print.R
```diff
@@ -73,13 +73,15 @@ print.gsplot <- function(x, ...){
...
         bgCol(bg.arg)
```
lindsayplatt: bgCol covers ticks :-1:

PR #365, opened 2016-06-30
updates for two issues

jordansread:


jordansread:
@lindsaycarr review?

lindsayplatt:
Should it always draw side 1 and 2? With your changes, it only draws 2,3 or 1,4 and not the other two axes. Was that the previous behavior?

``` r
gs <- gsplot() %>%
    points(1:10, 1:10, side=4)
gs
```

image

jordansread:
that should be the existing behavior. To test, try it against previous releases of `USGS-R/gsplot` as side=c(1,4) (since side=4 will break, hence the issue).

lindsayplatt:
You're correct! Merging now

jordansread:
note if you want all four, you have to be explicit:

``` r
gs <- gsplot() %>%
    points(1:10, 1:10, side=4) %>% axis(2) %>% axis(3)
gs
```

image

lindsayplatt:
RIght, I think what was really throwing me off is that frame.plot is missing still

Issue #350, opened 2016-06-22
where does box args reside (ex. frame.plot)


lindsayplatt:
@ldecicco-USGS's comment from #344 

As for framing (ie...whether or not to put a box around)...Maybe we could inject the plot.defaults at the first instance of gsplot(), and then only override them if the user does.

```
## Default S3 method:
plot(x, y = NULL, type = "p",  xlim = NULL, ylim = NULL,
     log = "", main = NULL, sub = NULL, xlab = NULL, ylab = NULL,
     ann = par("ann"), axes = TRUE, frame.plot = axes,
     panel.first = NULL, panel.last = NULL, asp = NA, ...)
```

That was my intent waaay back in the day. I can't figure out of that just got lost in the overhaul, or it's acting up.

jordansread:
putting this in `gs$global$config$frame.plot=TRUE`

lindsayplatt:
is that where asp and other plot.default args with no other home would live?

jordansread:
technically, `asp` is a view thing I think, so it should be in `gs$view.1.2$config$asp` following this pattern

jordansread:
but, Walker mentioned having the ability to bubble up those values to the global level so that they would apply across all views.

PR #369, opened 2016-06-30
frame.plot

jordansread:
we now have a box by default

``` r
gsplot() %>% points(1,2)
```

image

``` r
gsplot() %>% points(1,2, frame.plot = FALSE)
```

image

unlike `graphics::plot`, setting `axes=FALSE` does not also turn frame.plot to FALSE, so 

``` r
gsplot() %>% points(1,2, axes = FALSE)
```

image


R/gsplot-class.R
```diff
@@ -14,7 +14,7 @@ gsplot <- function(x = NULL, ...) UseMethod("gsplot")
 #' @rdname gsplot
 #' @export
 gsplot.default <- function(...) {
-  object <- gsplot(list())
+  object <- gsplot(list(global=list('config'=list(frame.plot=TRUE))))
```
jordansread, line 17: @jiwalker-usgs I'd like to hear where you think these defaults should live. Obviously, not here. But, I can't add them to default.yaml w/o hacks

jordansread:
forgot to build vignettes...that is coming in a commit shortly

lindsayplatt:
Fixing #350

Issue #207, opened 2015-08-14
allow reordering of gsplot list items


lindsayplatt:

ldecicco-USGS:
Can you give an example, like how you'd like to see it done?

jordansread:
I would think we should allow 1) reshuffling of the functions in each view, and 2) reshuffling of the view render order..others?

lindsayplatt:
Would we need that to be a specific gsplot ability? Because I have been doing that for AQCU after the gsplot object was created.

lindsayplatt:
- lowest usability, easiest implementation: passing in `where =` into the `after` argument in append `append(x, values, after = length(x))`
- where="last", but could change to "first"
- highest usability, hardest implementation: `z.index` changes order on render (1 = first rendered)
- use `reorder` function and assign id's and use those to reorder (default ids - `lines1`, `lines2`, etc). ids copied into auto legend. `reorder` is an extension to gsplot

ids (superuser) + where (simplest use case)

jordansread:
@lindsaycarr here is what I have so far:

``` r
gsplot() %>% points(1,2, cex=3.5, pch=0, lwd=3) %>% 
   points(1,2, pch=17, col='green', cex=4)
```

![image](https://cloud.githubusercontent.com/assets/2349007/16602309/afd8531c-42d8-11e6-9743-c5dc4c580170.png)

``` r
gsplot() %>% points(1,2, cex=3.5, pch=0, lwd=3) %>% 
   points(1,2, pch=17, col='green', cex=4, where='first')
```

![image](https://cloud.githubusercontent.com/assets/2349007/16602316/baa44224-42d8-11e6-9bb1-412d9826a128.png)

lindsayplatt:
Simple implementation is complete. Closing this issue and can revisit the z.index idea if it becomes a desired feature.

PR #378, opened 2016-07-05
adding "where" param

jordansread:
lets us do first or last (default is "last")

part of 
#207


jordansread:
Note that this is not reflected in the `legend` ordering right now. I am not very familiar with that. So if we like this, we should create an issue to also include first or last in the way we append the legend contents for auto legends

jordansread:
added the legend re-ordering 

``` r
gs = gsplot() %>% points(1:2,2:3, cex=c(2.5,1), pch=0, lwd=3, legend.name=c('points.1','points.2'), bg='cyan', xlim=c(0.5, 2.5), ylim=c(1.5,3.5)) %>% 
    points(1,2, pch=17, col='green', cex=4, where='first', legend.name='points.3', bg='dodgerblue') %>% legend()
gs
```

image

I tried a bunch of different combos of things, and it seems things are operating as expected.

Issue #381, opened 2016-07-06
Make view function for passing arguments through to views and sides


jiwalker-usgs:
Replaces #377 after discussion

PR #382, opened 2016-07-06
View filter

jordansread:
#381


jordansread:
for @lindsaycarr 's issue in `repgen`, instead of using the empty lines call, you could do

``` r
gs <- view(gsplot(), ylim=c(0, 8.5))
```

also, you can put anything in there that should end up in `option.args`, including `frame.plot` and `axes` etc.

Issue #343, opened 2016-05-03
draw_axis should share logic with draw_custom_grid


jordansread:
looks like these two functions have different ways of getting the ticks for axis/grid. Should make sure these are consistent.

lindsayplatt:
First: figure out what major issues are

(eventually removing draw functions - should be done in build so that object has everything it needs)

PR #384, opened 2016-07-06
Grid fix

lindsayplatt:


R/grid.R
```diff
@@ -65,13 +65,29 @@ draw_custom_grid <- function(object, view.name){
   grid.args <- remove_field(grid.args, "equilogs")
   at <- list()
   for (side.name in side.names){
-    usr.at <- axis_axTicks(object, as.side(side.name))
-    if (is.null(usr.at))
-      at[[side.name]] <- grid_axTicks(object, as.side(side.name))
-    else
-      at[[side.name]] <- usr.at
+    side.axis <- as.axis(side.name)
+    if(side.axis == "x")
```
ldecicco-USGS: I would love it if we could use curly brackets for all if else statements....I know we don't have an official required style...but it's so much easier to follow in my opinion.

lindsayplatt: tried to stick with the style this one had, but yes, I agree. I like curly brackets

R/grid.R
```diff
@@ -65,13 +65,32 @@ draw_custom_grid <- function(object, view.name){
...
+        at[[side.name]] <- grid_axTicks(object, as.side(side.name))
+      } else {
+        at[[side.name]] <- usr.at
+      }
+    } else {
+      plot.lims <- lim(object, as.side(side.name))
+      if(is.na(n.side)) {
+        at[[side.name]] <- NULL
+      } else {
+        at[[side.name]] <- seq.int(plot.lims[1], plot.lims[2], length.out = n.side + 1)
```
ldecicco-USGS, line 87: @lindsaycarr does this work for all types of data?

lindsayplatt, line 87: @ldecicco-USGS no, no it does not. Figured this was pretty incomplete. Thanks for pointing that out. Struggling with the best implementation of this function - is there a reason why `grid.gsplot` does not eventually call on `graphics::grid`? That's the pattern we follow for a lot of the other functions. Otherwise, we will be duplicating code. I know the graphics version already handles all of that

ldecicco-USGS, line 87: Yeah, grid was special because the base behavior of grid was pretty lame (not following ticks). I thought at one point both the fancyness of the new grid, and the nx, ny of old worked together, but I don't know when.

ldecicco-USGS, line 87: I don't have time to do this today, but I think we want to reverse engineer how nx,ny works in base and follow that (ie calling axTicks or something like that). That being said, we could merge this pull request, and make an issue that nx/ny doesn't work on non-integers if that would help for repgen

ldecicco-USGS, line 87: https://svn.r-project.org/R/trunk/src/library/graphics/R/grid.R

ldecicco-USGS, line 87: So, if a user specified `axis`, then the call to `grid` would be wrong (hence the re-write). That's the historical reason, but we do obviously still need nx/ny options

jordansread, line 87: @ldecicco-USGS I think (or I hope) the intent is covered in https://github.com/USGS-R/gsplot/issues/343 ?

lindsayplatt, line 87: So we never supported nx/ny - which is weird because I've been using that in repgen. Hmmmm apparently it hasn't been doing what I needed it to do? Anyways, I would like to bump up #343 if that's possible. I can work on it, but @jread-usgs can you add a little more detail?

jordansread, line 87: yes, I will add detail

jordansread:
replaced with #386

PR #388, opened 2016-07-08
Legend fix

### LINKED ISSUE #387: legend not adding legend.name if created first
``` r
gs <- gsplot() %>% 
    legend() %>% 
    points(1,2, legend.name="stuff")
names(gs$legend$legend.auto)
[1] "location"      "legend_offset" "draw"   
```

image

lindsayplatt:
#387


lindsayplatt:
@jread-usgs @jiwalker-usgs @ldecicco-USGS looking for peer review on this if you have time

man/get_legend_arg_names.Rd
```diff
@@ -0,0 +1,19 @@
...
+% Please edit documentation in R/modify_legend.R
+\name{get_legend_arg_names}
+\alias{get_legend_arg_names}
+\title{get vector of legend arguments - overall or for each entry}
+\usage{
+get_legend_arg_names(overall = FALSE, indiv = FALSE,
+  names.args = names(formals(graphics::legend)))
+}
+\arguments{
+\item{overall}{logical indicating whether overall legend arguments should be returned}
```
jordansread: Can you define what the difference between overall args are vs non-overall?

R/modify_legend.R
```diff
@@ -208,3 +214,32 @@ modify_legend <- function(object, location="topright", legend_offset=0.3, draw=F
...
 }
+
+#' get vector of legend arguments - overall or for each entry
+#' 
+#' @param overall logical indicating whether overall legend arguments should be returned. These arguments impact 
+#' the legend style (e.g. title, box.lwd, location, etc)
+#' @param indiv logical indiciating whether arguments applicable to each legend entry should be returned. These are 
+#' the arguments that apply to each individual entry within the legend (e.g. lty, lwd, pch, text.font, etc)
+#' @keywords internal
+get_legend_arg_names <- function(overall = FALSE, indiv = FALSE, names.args = names(formals(graphics::legend))){
```
jordansread: thinking of different things to do for the `names(formals...)` etc, because I think we want to have the handling of that be consistent for the package. 

This 

``` r
formal_names('legend')
```

would be preferred, but you can't drop it in as the default because `formal_names` is a `gsplot:::` function. Perhaps use default of NULL which means all or `"all"` as default which means call `formal_names('legend')` within the function?

lindsayplatt: well I don't think it needs to be an argument now that I am looking again. The idea was that `names.args` is just always the formals of `legend`, so that isn't a good use of a function argument. It should really just go inside the function

R/modify_legend.R
```diff
@@ -149,30 +149,36 @@ combine_legend_args <- function(object, new.legend.args, ..., where){
...
   legend.args <- object[['legend']][['legend.auto']]
   
-  orderedParams <- new.legend.args[match(names(legend.args), names(new.legend.args))]
-  for (j in seq_along(legend.args)) {
-    if (where == 'first'){
-      legend.args[[j]] <- c(orderedParams[[j]], legend.args[[j]])  
-    } else {
-      legend.args[[j]] <- c(legend.args[[j]], orderedParams[[j]])  
-    }
+  which.overall.args <- which(names(legend.args) %in% get_legend_arg_names(overall = TRUE))
```
jordansread: kind of prefer returning a logical vector here (instead of `which`) and then using `[!overall.args]` below

lindsayplatt: any reason?

jordansread: Because you don't use the integers that come back from this: `names(legend.args) %in% get_legend_arg_names(overall = TRUE)`, so it kind of just adds an additional function call (to `which`) that isn't necessary.

jordansread:
@lindsaycarr looks like it is failing a legend test. Otherwise, looks good.

PR #390, opened 2016-07-08
change order of print

lindsayplatt:


lindsayplatt:
@jread-usgs review and merge?

jordansread:
@lindsaycarr  can you run `devtools::build_vignettes()` and knit the readme as part of this PR? Just want to see what all it changes visually.

R/legend.R
```diff
@@ -103,6 +103,7 @@ draw_legend <- function(gsplot) {
         legend <- lapply(legend, function(x) {unname(sapply(x, function(x) {eval(x)}))})
         # clean out arguments not allowed by legend 
         legend <- legend[na.omit(match(names(default.args), names(legend)))]
+        par(usr=view.usr)
```
jordansread: do you reset the par after this function uses `usr`?

jordansread:
Looks like we still have an issue with the legend location? Or do you need to re-knit the vignettes? 
image

R/legend.R
lindsayplatt: no I do not, but when we had the print order with `draw_axis` and then `draw_custom_grid`, the `usr` remained these values throughout `draw_legend`. So actually, the better way to handle this would be to change `usr` in `print`. Then I wouldn't need to alter `draw_legend` at all

lindsayplatt:
I need to re-knit

lindsayplatt:
Wanted to make sure changing par here wasn't too hacky before doing that

jordansread:
I think this level of hacky-ness is on par with the existing function, so I don't see a problem. Eventually, we will want to overhaul how this is done to be more robust and fit with the other rendering pattern we are converging on.

lindsayplatt:
@jread-usgs ok, but that won't affect gsplot stability since it is behind-the-scenes, correct? Hoping to release new version later this week

jordansread:
correct

Issue #395, opened 2016-07-11
lost date formatting/handling


lindsayplatt:
dates are coerced to numeric now
- [ ] POSIXct
- [ ] Date
- [ ] test for POSIXct
- [ ] test for Date

``` r
some_dates <- seq(as.POSIXct("2010-10-01"), as.POSIXct("2011-09-30"), by="month")
gs <- gsplot() %>% 
  points(some_dates, 1:12)
gs
```

![image](https://cloud.githubusercontent.com/assets/13220910/16740084/6ef270a2-476b-11e6-9dd7-21148040c24b.png)

jordansread:
weird part is that `xlim` is correct:

``` r
xlim(gs)
$side.1
[1] "2010-10-01 CDT" "2011-09-01 CDT"
```

is the bug in `draw_axis`?

jordansread:
Note what `draw_custom_grid` does, which draw_axis does not do (see #343 ?)

``` r
if(is.numeric(lim)){
    at <- axTicks(side)
  } else{
    fun <- getFromNamespace(paste0('axis.',class(lim)[1L]),'graphics')
    at <- fun(side, x = lim, lwd=0, lwd.ticks=0, labels = FALSE)
  }
```

but that is for `at`, not for the label

lindsayplatt:
see #396

PR #396, opened 2016-07-11
draw_axis handling different data types

lindsayplatt:
fixing #395 

sharing some helper functions with `draw_custom_grid`, but not quite to the extent #343 is talking about


R/axis.R
```diff
@@ -88,14 +88,22 @@ axis.gsplot <- function(object, ..., n.minor=0, tcl.minor=0.15, reverse=NULL) {
   
 }
 
-draw_axis <- function(axis.args){
+draw_axis <- function(object, side.args){
```
jordansread: aren't `side.args` redundant w/ `object`? How about just `object` and `side` or `side.name`?

R/print.R
```diff
@@ -71,7 +71,7 @@ print.gsplot <- function(x, ...){
     old.par <- par(x[[side.name]]$par)
     set_frame(views, side)
     if(x[[side.name]][['axes']] | x[[side.name]][['usr.axes']]){
-      draw_axis(x[[side.name]][['axis']])
+      draw_axis(x, x[[side.name]])
```
jordansread: I'd suggest 

``` r
draw_axis(x, side.name)
```

R/axis.R
```diff
@@ -88,14 +88,23 @@ axis.gsplot <- function(object, ..., n.minor=0, tcl.minor=0.15, reverse=NULL) {
...
-draw_axis <- function(axis.args){
+draw_axis <- function(object, side.name){
+  axis.args <- object[[side.name]][['axis']]
+  side.lim <- object[[side.name]][['lim']]
+
+  if(all(is.na(side.lim)) || is.numeric(side.lim)){
+    fun.name <- 'axis'
+  } else {
+    fun.name <- paste0('axis.', class(side.lim)[1L])
+    axis.args$at <- get_axTicks(object, as.side(side.name))
```
jordansread: can you remove this line? What does it do? I think you want to let the function `axis.class` figure out the `at`

lindsayplatt: as.POSIXct and axis.Date don't allow `at` to be NULL, so I had to define it and then pass it in

lindsayplatt: it will just return `Error in as.Date(x) : argument "x" is missing, with no default`

jordansread: ahh, ok. thanks

jordansread: Looks like `axis.args$at` doesn't handle

``` r
gs <- gsplot() %>% 
     points(some_dates, 1:12) %>% 
     axis(1)
gs
```

lindsayplatt: in what way?

jordansread: geez, sorry. I flipped branches too many times trying to track this down. It does handle that as expected. 

I am completely confused why you can give `axis` all those `at` values and it still only renders the two extremes. This issue has me totally stumped.

lindsayplatt:
using `pretty()` and everything does match base!

lindsayplatt:
Base and gsplot are matching now

**Base**

``` r
some_dates <- seq(as.POSIXct("2010-10-01"), as.POSIXct("2011-09-30"), by="month")
plot(some_dates, 1:12)
```

image

**gsplot**

``` r
gs <- gsplot() %>% 
  points(some_dates, 1:12)
gs
```

image

lindsayplatt:
Another example:

**Base**

``` r
library(dataRetrieval)
siteNumber <- '04085427'
startDate <- '2012-01-01'
endDate <- '2012-06-30'
pCode <- '00060'
dv <- readNWISdv(siteNumber,pCode, startDate, endDate)

plot(dv$Date, dv$X_00060_00003)
```

image

**gsplot**

``` r
gsplot() %>%
  points(dv$Date, dv$X_00060_00003)
```

image

lindsayplatt:
### Test for this is in the README

see _Compatibility with base plotting_ section

``` r
date_vector <- seq(as.Date("2010-10-01"), as.Date("2011-09-30"), by="months")
gs <- gsplot() %>% 
  points(date_vector, 1:12)
gs

points(as.Date("2010-11-15"),2.5)
```

What the plot would look like before this PR:
image

After this PR:
image

R/grid.R
```diff
@@ -96,8 +96,7 @@ grid_axTicks <- function(object, side){
   if(is.numeric(lim)){
     at <- axTicks(side)
   } else{
-    fun <- getFromNamespace(paste0('axis.',class(lim)[1L]),'graphics')
-    at <- fun(side, x = lim, lwd=0, lwd.ticks=0, labels = FALSE)
+    at <- Axis(side = side, x = pretty(lim), lwd=0, lwd.ticks=0, labels = FALSE)
```
jordansread, line 99: Ahh, so this is the key change?

lindsayplatt, line 99: yep - needed `pretty()` so that it wasn't just passing two values. Then switched to using `Axis` because that internally handles all of the `axis.[class]` stuff

Issue #391, opened 2016-07-08
Custom configs bleed over if default config 


ldecicco-USGS:
If a config file has something that is not defined in default:
superFat.yaml:

```
lwd: 3
lines: 
  lwd: 3
abline:
  col: "grey"
  lty: 1
  lwd: 1
```

Then:

``` r
df <- data.frame(x = 1:10, y=1:10, z = seq(2,20,2))
loadConfig("D:/LADData/RCode/gsplot/superFat.yaml")
gsp <- gsplot() %>%
  lines(df$x, df$y, col="red", legend.name = "points") %>%
  lines(df$x, df$z, legend.name = "lines") %>%
  legend()
gsp
```

Makes this:
![2016-07-08_14-58-13](https://cloud.githubusercontent.com/assets/1105215/16699905/94629246-451c-11e6-926d-f715af1e722d.png)

Then do:

``` r
loadConfig()
gsp <- gsplot() %>%
  lines(df$x, df$y, col="red", legend.name = "points") %>%
  lines(df$x, df$z, legend.name = "lines") %>%
  legend()
gsp
```

and you still get:
![2016-07-08_15-00-21](https://cloud.githubusercontent.com/assets/1105215/16699936/c4e91304-451c-11e6-9d84-2d1bc04ebe5f.png)

ldecicco-USGS:
So, we need to make some sort of "clear all" function

PR #400, opened 2016-08-04
Fix par bleeding

ldecicco-USGS:
As described in #391 

So now we add a new list of original par (as defined on the loading of the package) to our `gsconfig` env , and add that to the custom gsconfig$options when `loadConfig` is called. I therefore had to change some tests. 

@jread-usgs @jiwalker-usgs and/or @lindsaycarr have a look.


R/config.R
```diff
@@ -1,10 +1,12 @@
 gsconfig <- new.env(parent = emptyenv())
+gsconfig$orignial.par <- par(no.readonly = TRUE)
```
lindsayplatt: just so this doesn't mix us up later - the n and the i are mixed up here: original

ldecicco-USGS: that sounds like my coding!

R/config.R
```diff
@@ -21,10 +23,14 @@ loadConfig = function(filename) {
   }
 
   graphTemplate <- yaml.load_file(filename)
-
+  origPar <- gsconfig$orignial.par
```
lindsayplatt: spelling here too

lindsayplatt:
in the updated figs, some of the stuff looks like it's getting cutoff now

R/config.R
ldecicco-USGS: I'm also wondering if we want to do a dev.new right before catching the par to truly get R's defaults, or is there some off-chance that someone will have set par assume it's going to persist?

R/config.R
ldecicco-USGS: well, if it was only misspelled in one location, it wouldn't work

ldecicco-USGS:
I took out my changes, rebuilt the Readme, and the newly created figures (for the readme) look the same as with my changes I'm proposing here (ie...these changes aren't causing the altered figures). @lindsaycarr could you verify that your fork's readme (assuming it's current with USGS-R) produces the graphs without changes? If you don't see altered figures, could you pull this PR locally and see if you produce the same readme figures that I now?

What's weird and interesting is that the graphs I'm pushing here look fine when done in the R console...but different when created via knitr. So that may be a separate issue.

lindsayplatt:
So when I knit the README on my local version (updated with most recent upstream), none of the files changed. I pulled down your branch, and knit without building and reloading and the figures changed but back to how I would expect them to look. Then I built and reloaded and none of the figures show up in the Git tab, but they look different than I would expect them to.

ldecicco-USGS:
huh? "Then I built and reloaded and none of the figures show up in the Git tab, but they look different than I would expect them to."

PR #401, opened 2016-08-15
Getting knitr to work again

ldecicco-USGS:


R/config.R
```diff
@@ -22,7 +22,8 @@ loadConfig = function(filename) {
   }
 
   graphTemplate <- yaml.load_file(filename)
-  if(.Device != "null device"){
+
+  if(!(.Device %in% c("null device","png"))){
```
ldecicco-USGS, line 26: OK, so seemed like a great/easy/straightforward solution. Until I discovered the graphs weren't working right in `knitr`. 

I'm not super-happy with this hack. The alternative I think is comparing original.par with new par. That being said....this line does work in all cases I can envision. The situation where it could be problematic that I want to test is changing the config file mid-knitr doc.

Issue #392, opened 2016-07-08
Put a config argument in gsplot


ldecicco-USGS:
To load a config file just for a particular object

lindsayplatt:
Closed by #416

Issue #254, opened 2015-09-28
helper function for temporarily changing pars


lindsayplatt:
set and reset pars easily

jordansread:
Anyone have any ideas for how this should be done? What do you revert to?
- a pkg.env variable that stores the user's par pre-render?
- another pkg.env var that stores a view's par (or pars for all views?)
- or just a pkg.env var that stores the incoming par whenever `set_par(...)` is used (just an example call) 
- or `set_par` just returns the current (`read.only=F`) par before it is changed, but with `invisible`??

Just looking for input on the desired workflow. I find myself wanting to set and revert to the previous state. How about y'all?

I think the last option is the easiest, and would look like this:

``` r
set_par(mar=c(0,0,0,0))
# changes mar to 0,0,0,0
old.par = set_par(mar=c(1,2,0,0))
# returns a par list, w/ mar=c(0,0,0,0)
set_par(old.par) 
# reverts to mar=c(0,0,0,0)
```

lindsayplatt:
@jread-usgs is this a For Later task to you, or needed for MVP?

lindsayplatt:
Discussed and determined not needed for "go-live". Could add later.

Might be useful internally to gsplot to help with us constantly setting and resetting pars.

PR #402, opened 2016-08-15
Config branch...

ldecicco-USGS:
I had been using the dev.off to fix the bleeding par, but that had an unintended consequence when kniting. This PR instead goes back to the original scheme where we'd save the original par in the gsconfig, then reset all of those pars if someone changes the config (either with loadConfig or in the call to the gsplot object). This somewhat alters all the figures because now the background tends to be translucent instead of white. I actually think that's better, but would like to hear more discussion.


R/config.R
```diff
@@ -1,4 +1,5 @@
 gsconfig <- new.env(parent = emptyenv())
+gsconfig$orignial.par <- par(no.readonly = TRUE)
```
lindsayplatt: @ldecicco-USGS "original"

R/config.R
```diff
@@ -39,14 +63,15 @@ loadConfig = function(filename) {
 #' @param type string of gsplot config object to retrieve
 #' @param ... additional configuration to override what is pulled from config
 #' @param persist logical of whether to persist overrides to config
+#' @param custom.config logical of whether to use default global (TRUE) or a config set for only one gsplot object
```
lindsayplatt: custom.config=TRUE would be for the config set, FALSE would indicate defaults

R/config.R
ldecicco-USGS: HAHAHAHA

ldecicco-USGS: yeah, learned how to wipe out a branch and this is what happens

R/legend.R
```diff
@@ -99,8 +99,9 @@ draw_legend <- function(gsplot) {
         #set bg so that fill/border/etc args are correct, then evaluate any quoted list items
         if (any(names(legend) %in% c("bg"))) {
           par(bg=legend$bg)
+        } else {
+          par(bg = ifelse(par('bg') == "transparent", "#FFFFFF", par('bg')))
         }
-        legend <- lapply(legend, function(x) {unname(sapply(x, function(x) {eval(x)}))})
```
lindsayplatt, line 103: I think we still need this - it evaluates anything in legend that we set with `quote()`

R/legend.R
```diff
@@ -99,8 +99,9 @@ draw_legend <- function(gsplot) {
         #set bg so that fill/border/etc args are correct, then evaluate any quoted list items
         if (any(names(legend) %in% c("bg"))) {
           par(bg=legend$bg)
+        } else {
+          par(bg = ifelse(par('bg') == "transparent", "#FFFFFF", par('bg')))
```
lindsayplatt, line 103: :+1: I really did not know that "transparent" was an option!

R/legend.R
ldecicco-USGS, line 103: agreed, I think that's a pasting error on my part, I certainly never planned to delete that line.

lindsayplatt:
@ldecicco-USGS I'm on your branch (and built/reloaded), but this is still causing an issue. It's like the way legend handles par is messing it up because it looks ok when I exclude `legend.name`. However, not sure why the `axis()` call is being ignored for changing `side.2$lim`.

``` r
g1 <- gsplot() %>% 
  lines(1:9, seq(1,3, by=0.25), 

  # formatting y axes
  axis(side=2, at=1:10) 

layout(matrix(1:2, 1, 2))
g1
g1
```

image

ldecicco-USGS:
I did not test out my branch at all with `layout`...I'm not sure what happens with `par` there. 

I totally agree that `legend` is mucking up nice `par` behaviors. I've noticed this in the past too. It's the only function that directly changes `par`, and then unchanges it back at the end. I'm guessing that's the root of the problem.

I'm guessing what you are seeing there is something I've been concerned about for a little while now. Because we are changing `par` in the building up of the gsplot object....it's possible to get 2 different printed outputs using the same object if `par` gets changed from one print to the next. I'm guessing that's what's going on here. I'd have to explore what `layout` does with `par` to know what's going on here and how we'd want to deal with it in the longer term.

lindsayplatt:
It happens when you just print the plot twice in a row as well, layout was just easier to show them side by side

lindsayplatt:
It only changes par when draw_legend is called, which doesn't happen here because I never have the legend drawn. I only added legend.name, but never used legend(). So I think something else is at play here

lindsayplatt:
Actually, calling legend() just made that issue stop...what is happening?!

``` r
g1 <- gsplot() %>% 
  lines(1:9, seq(1,3, by=0.25), 

  # formatting y axes
  axis(side=2, at=1:10) %>% legend()

layout(matrix(1:2, 1, 2))
g1
g1
```

image

ldecicco-USGS:
I'm just saying:
1. I think we need to re-think how we deal with draw_legend _in general_ I know this is not the case that you are talking about right here and now
2. This particular case is different, and I'm looking into it (personally, on my own time at home tonight)

ldecicco-USGS:
Here's how I'd start doing some troubleshooting for par things:

```
g1 <- gsplot() %>% 
  lines(1:9, seq(1,3, by=0.25), 
  axis(side=2, at=1:10) 
x <- par(no.readonly = TRUE)
g1
y <- par(no.readonly = TRUE)
all.equal(x, y)
 [1] "Component “mgp”: Mean relative difference: 0.55"      
 [2] "Component “pin”: Mean relative difference: 0.1922864" 
 [3] "Component “plt”: Mean relative difference: 0.1891681" 
 [4] "Component “tcl”: Mean relative difference: 1.6"       
 [5] "Component “usr”: Mean relative difference: 6"         
 [6] "Component “xaxp”: Mean relative difference: 1.833333" 
 [7] "Component “xaxs”: 1 string mismatch"                  
 [8] "Component “xpd”: 1 element mismatch"                  
 [9] "Component “yaxp”: Mean relative difference: 0.6666667"
[10] "Component “yaxs”: 1 string mismatch"          
```

ldecicco-USGS:
The difference between the 2 graphs in `par` seems to be xpd...

```
g1 <- gsplot() %>% 
  lines(1:9, seq(1,3, by=0.25), 
  axis(side=2, at=1:10) %>% legend()
x <- par(no.readonly = TRUE)
g1
y <- par(no.readonly = TRUE)
all.equal(x, y)
[1] "Component “mgp”: Mean relative difference: 0.55"      
[2] "Component “pin”: Mean relative difference: 0.1922864" 
[3] "Component “plt”: Mean relative difference: 0.1891681" 
[4] "Component “tcl”: Mean relative difference: 1.6"       
[5] "Component “usr”: Mean relative difference: 6"         
[6] "Component “xaxp”: Mean relative difference: 1.833333" 
[7] "Component “xaxs”: 1 string mismatch"                  
[8] "Component “yaxp”: Mean relative difference: 0.6666667"
[9] "Component “yaxs”: 1 string mismatch"        
```

lindsayplatt:
Oh and having xpd be different would explain all of that. Adding this fixes the issue, so now I can figure out where that's being changed.

``` r
g1 <- gsplot(xpd=FALSE) %>% 
  lines(1:9, seq(1,3, by=0.25), 

  # formatting y axes
  axis(side=2, at=1:10) 

layout(matrix(1:2, 1, 2))
g1
g1
```

lindsayplatt:
should we up to 0.7.1 when this merges??

ldecicco-USGS:
Yeah

lindsayplatt:
I tested this with repgen and the only thing I noticed that was different was a legend was being cutoff. I'm trying to recreate a simple example.

lindsayplatt:
Something with this branch seems to be ignoring `par`.

``` r
par(mar=c(5,4,10,2))

g1 <- gsplot() %>% 
  points(1:10, 1:10, pch=20, legend.name="first points") %>% 
  lines(4:1, 4:1, legend.name="first line") %>% 
  points(c(3,7,4), c(9,3,6), pch=20, col="black", legend.name="second points") %>% 
  legend(location="above")
g1
```

### v0.7.0

image

### this branch

image

ldecicco-USGS:
I can fix that...maybe by the end of the day...silly mistake

lindsayplatt:
Oh ok cool! Think that might also fix #405 ??

ldecicco-USGS:
Forgot about this...will try to do soon

lindsayplatt:
I was thinking we could take a look tomorrow at the sprint planning meeting as well

R/add_to_view.R
```diff
@@ -70,14 +71,14 @@ add_to_view <- function(object, call.args, side, where){
 #' gsplot:::filter_arguments('points', x=1:5, y=1:5, xlim=c(0,10), ylim=c(0,10), 
 #'                callouts(labels=c(rep(NA, 4), "oh")))$extracted.args
 #' @keywords internal
-filter_arguments <- function(fun.name, ...){
+filter_arguments <- function(fun.name, ..., custom.config = FALSE){
```
jordansread, line 74: was thinking that default would be `custom.config = NULL` since that is what you would get if 
`custom.config = object[["global"]][["config"]][["config.file"]]` didn't exist. Maybe I missed it, but where do you test for `!custom.config` or where is that used?

lindsayplatt:
@ldecicco-USGS do you remember what we concluded about this? Is there more work to do?

I see that it potentially fixes #405, #392, #130

lindsayplatt:
par issues: https://github.com/USGS-R/gsplot/issues/254

pass config a list or a filename

ldecicco-USGS:
I'm just going to start over

Issue #408, opened 2016-10-24
overlap/overwrite axis calls


lindsayplatt:
``` r
gs <- gsplot() %>% 
    points(0:1,0:1) %>% 
    axis(side=1, at=c(0.5,1)) %>% 
    axis(side=1, at=c(0.25, 0.75))
gs
```

Should plot have 0.25, 0.5, 0.75, and 1 ticks to match base behavior?

![image](https://cloud.githubusercontent.com/assets/13220910/19658263/ef020bb2-99ec-11e6-8251-b6c3ee9a6dc4.png)

Suggestion:

``` r
gs <- gsplot() %>% 
    points(0:1,0:1) %>% 
    axis(side=1, at=c(0.5,1)) %>% 
    axis(side=1, at=c(0.25, 0.75), append=TRUE)
gs
```
- always assume axis call exists (default) for append/replace
  to not draw axis lines: use lwd=0 and lwd.ticks=1 ??

lindsayplatt:
closed by #410

PR #410, opened 2016-10-25
support for several axis calls on same side

jordansread:


R/axis.R
```diff
@@ -76,9 +77,25 @@ axis.gsplot <- function(object, ..., n.minor=0, tcl.minor=0.15, reverse=NULL) {
...
+      last.axis <- tail(which.axis, 1) # get the last one
+      object[[as.side_name(side)]] <- append(object[[as.side_name(side)]], 
+                                             list('axis' = set_args('axis', side=side, package='graphics')), 
+                                             after = last.axis)
+      object[[as.side_name(side)]][[last.axis+1]] <- append_replace(object[[as.side_name(side)]][[last.axis+1]], user_args[[fun.name]])
+    } else {
+      # remove
+      if (length(which.axis) > 1){
+        # remove all axis functions other than the first one
+        object[[as.side_name(side)]] <- object[[as.side_name(side)]][-which.axis[!which.axis %in% head(which.axis, 1)]]
```
lindsayplatt, line 93: I was thinking you'd want to remove everything but the last one

R/axis.R
```diff
@@ -89,6 +106,16 @@ axis.gsplot <- function(object, ..., n.minor=0, tcl.minor=0.15, reverse=NULL) {
 }
 
 draw_axis <- function(object, side.name){
+  # method isn't made for multiple axis calls
+  which.axis <- which(names(object[[side.name]]) == 'axis')
+  if (length(which.axis) > 1){
+    for (axis.i in which.axis){
+      tmp <- object
```
lindsayplatt, line 113: what's the purpose of creating tmp?

tests/testthat/tests-axis.R
```diff
@@ -46,6 +46,40 @@ test_that("axis reverse",{
...
+    axis(side=1, at=c(0.45, 0.55), append=TRUE)
+  
+  expect_equal(sum(names(gs$side.1) == 'axis'), 3)
+  gs <- gsplot() %>% 
+    points(0:1,0:1) %>% 
+    axis(side=1, at=c(0.5,1)) %>% 
+    axis(side=1, at=c(0.25, 0.75), append=TRUE) %>% 
+    axis(side=1, at=c(0.45, 0.55), append=TRUE) %>% 
+    axis(side=1, at=c(0.33))
+  expect_equal(sum(names(gs$side.1) == 'axis'), 1)
```
lindsayplatt, line 72: any way to add a test for making sure it's keeping the expected axis and dropping the others?

jordansread: good idea @lindsaycarr I will add that

R/axis.R
jordansread: this was kind of a hack to deal with multiple axis function calls in one object. tmp is the same object, but w/o the other axis calls, that way I can pass tmp back into the same function and it will be treated the same way a single axis object is. Make sense? This function in general needs some attention, and I just added some new junk to it.

R/axis.R
jordansread: hmm...we should discuss the expected behavior for this.

jordansread: @ldecicco-USGS and @jiwalker-usgs do you have thoughts on this?

jordansread: My reasoning was this:
using the `append` arg to get multiple axis calls in there means appending them to the original call, which is the default set-up when the side is created. When you then set `append = FALSE` (what it is checking for here), those appended axis calls are wiped out. 

I can see your point too, where you would want to treat it like append was FALSE the whole time and use the last, but that isn't the same behavior either.

For example:

``` r
gs <- gsplot() %>% 
    points(0:1,0:1) %>% 
    axis(side=1, at=c(0.5,1), labels=FALSE) %>% 
    axis(side=1, at=c(0.25, 0.75), line = 2, append=TRUE) %>% 
    axis(side=1, at=c(0.45, 0.55), lwd = 2, append=TRUE) %>% 
    axis(side=1, at=c(0.33))
```

the last call uses append = FALSE, so what does that mean? Ignore all of the other TRUE calls and treat them as FALSE (collapsing all of their arguments into one), or does it mean remove all appended `axis` calls and build off of the default (what it currently does)?

lindsayplatt: I would think in that example the last axis call replaces everything above it since it does not have append. 

``` r
gs <- gsplot() %>% 
    points(0:1, 0:1) %>% 
    axis(side=1, at=c(0.5, 1)) %>% 
    axis(side=1, at=c(0.25, 0.75), line = 2, append=TRUE) %>% 
    axis(side=1, at=c(0.45, 0.55), lwd = 2, append=TRUE) %>% 
    axis(side=1, at=c(0.33))
gs
```

Creates
image

R/axis.R
lindsayplatt: Hmmm I think I understand, but isn't tmp changing each time from the new axis on the next line?

R/axis.R
jordansread: yes, that is what happens now with this PR

R/axis.R
jordansread: the next line just has it remove all of the other axis calls in tmp. This loops through the axis calls and tmp is re-created each time based on the original object.

R/axis.R
lindsayplatt: hmmm I guess your comment says " remove all axis functions other than the first one" and that makes me think that everything except `axis(side=1, at=c(0.5, 1))` is removed. Nevermind if it's removing all except `axis(side=1, at=c(0.33))`!

R/axis.R
jordansread: the `object[[as.side_name(side)]][['axis']] <- append_replace(object[[as.side_name(side)]][['axis']], user_args[[fun.name]])` append_replace makes sure the new arguments override the existing ones.

tests/testthat/tests-axis.R
jordansread: added it

jordansread: @lindsaycarr can you merge if you think this is good to go?

lindsayplatt:
fixing #408

Issue #322, opened 2016-04-18
override bug with data.frame


jordansread:
Our rule to still fully support the masked `graphics` functions is failing in the data.frame case (probably others?)

``` r
# this works as expected
graphics::plot(1,3)
graphics::points(data.frame(1,2), col='red')

# this should have the same behavior, but errors instead:
plot(1,3)
points(data.frame(1,2), col='red')
 Error in xy.coords(x, y) : argument "x" is missing, with no default 
```

lindsayplatt:
y labels not showing anymore

jordansread:
@lindsaycarr can you provide more detail for this?

lindsayplatt:
see my comment in #411 on Nov 7

jordansread:
thanks

jordansread:
Confused on this one. Working from my `data_frame` branch:

``` r
gs <- gsplot() %>% points(1,2)
setdiff(gs$side.2$axis, gs$side.1$axis)
[[1]]
[1] 2
# so not different other than the side arg in axis
```

And the other args are the same other than limits (between sides 2 and 3) 

So I am thinking this issue must be in `print` or `draw_axis`. 

Same thing happening here https://github.com/USGS-R/gsplot/issues/414#issuecomment-261083821 @lindsaycarr ?

lindsayplatt:
@jread-usgs - @jiwalker-usgs and I were coming across this same axis labels disappearing issue yesterday when working in print. See my [last comment](https://github.com/USGS-R/gsplot/issues/414#issuecomment-261083821) on #414. We were stumped.

lindsayplatt:
@jread-usgs I've been debugging this for awhile, and replacing the if-else section with a method `params <- append_params(object, params)` is what causes the labels to disappear. I haven't been able to get much farther past identifying what causes the issue.

lindsayplatt:
That disappearing labels thing seemed unrelated to this now that we fixed it for #414. Maybe we should retry the fix that originally existed for this? Currently trying to figure out how to pull it down and test.

lindsayplatt:
If this is just for allowing regular graphics plot stuff to work, this branch fixes it (https://github.com/lindsaycarr/gsplot/commit/cae685c5ff568a076d07b175a5103aecaf52dd29). It does not allow gsplot to use data frames.

```r
gs <- gsplot() %>% 
    points(data.frame(1,2))
Error in is.finite(side.vals) : 
  default method not implemented for type 'list'
```

PR #411, opened 2016-10-25
Data frame

jordansread:
for #322


jordansread:
whoops, looks like there is something in the examples that isn't in tests.

ldecicco-USGS:
Is this ready to merge?

jordansread:
I think so..?

lindsayplatt:
@jread-usgs looks like this broke axis labeling, but only seems to be for side 2 and 4:

``` r
gs <- gsplot() %>% points(1,2)
gs
```

image

jordansread:
@lindsaycarr we should revert this then, right?

lindsayplatt:
I'd say yes...how exactly do we do that?

R/function_args.R
```diff
@@ -14,18 +14,15 @@ function_args <- function(package, name, object, ..., use.default=paste0(name,'.
   params <- list(...)
   
   if (!missing(object)) {
-    if (!is.null(names(object)))
-      params <- append(object, params)
-    else {
-      params <- append(list(object), params)
-    }
+    params <- append_params(object, params)
```
lindsayplatt, line 17: just isolated this as the issue that caused the labels to disappear.

jordansread: ahhh

R/axis.R
```diff
@@ -63,8 +63,7 @@ axis <- function(object, ...) {
   override("graphics", "axis", object, ...)
 }
 
-#' @importFrom utils head
-axis.gsplot <- function(object, ..., n.minor=0, tcl.minor=0.15, reverse=NULL, append=FALSE) {
+axis.gsplot <- function(object, ..., n.minor=0, tcl.minor=0.15, reverse=NULL) {
```
lindsayplatt, line 66: @jread-usgs why did we delete append=FALSE?

jordansread:
I don't know....Maybe that happened accidentally w a merge? Let's track down the commit and figure out why

Issue #285, opened 2015-11-25
When x values come in as POSIX, the xlim should have the same timezone


jordansread:
dates! :fire: 

Right now the xlim comes back w/ the user's timezone, even if the vals coming in to x have "UTC" or something else specified.

jordansread:
version `‘0.6.0’` (same as in `‘0.5.4’`)

``` r
x.date <- seq(as.POSIXct("2000-01-01", tz = 'UTC'), by = "month", length.out = 10)
x.date
 [1] "2000-01-01 UTC" "2000-02-01 UTC" "2000-03-01 UTC" "2000-04-01 UTC" "2000-05-01 UTC" "2000-06-01 UTC" "2000-07-01 UTC" "2000-08-01 UTC" "2000-09-01 UTC" "2000-10-01 UTC"
gs <- gsplot() %>%
       points(x.date,1:10) %>%
       lines(6:14,6:14,side=c(3,4)) %>%
       grid(side=c(3,4))

xlim(gs)
$side.1
[1] "1999-12-31 18:00:00 CST" "2000-09-30 19:00:00 CDT"

$side.3
[1]  6 14
```

lindsayplatt:
culprit: `strip_pts`
why: `out <- c()` and then `out <- append(out, v.vals)` where `v.vals` is the vector of dates. This makes v.vals a number and in the users timezone.

``` r
v.vals <- seq(as.POSIXct("2000-01-01", tz = 'UTC'), by = "month", length.out = 10)

out <- c()
append(out, v.vals)
 [1] "1999-12-31 18:00:00 CST" "2000-01-31 18:00:00 CST" "2000-02-29 18:00:00 CST" "2000-03-31 18:00:00 CST"
 [5] "2000-04-30 19:00:00 CDT" "2000-05-31 19:00:00 CDT" "2000-06-30 19:00:00 CDT" "2000-07-31 19:00:00 CDT"
 [9] "2000-08-31 19:00:00 CDT" "2000-09-30 19:00:00 CDT"

out <- c()
c(out, v.vals
[1] 946684800 949363200 951868800 954547200 957139200 959817600 962409600 965088000 967766400 970358400

# maybe do something like this: 
if(is.null(out)){
    out <- v.vals
} else {
    out <- append(out, v.vals)
}

out
[...]

PR #413, opened 2016-11-07
append without losing timezones in strip_pts

lindsayplatt:
Working on #285

### DIFF
--- a/R/utils-list.R
+++ b/R/utils-list.R
@@ -34,12 +34,7 @@ strip_pts <- function(list, param){
   for (v in param){
     if (v %in% names(list) && !inherits(list[[v]], c('function','formula'))) {
       v.vals <- list[[v]]
-      # append to out without losing timezones when it is a date/POSIX
-      if(is.null(out)){
-        out <- v.vals
-      } else {
-        out <- append(out, v.vals)
-      }
+      out <- append_keepclass(out, v.vals)
       out.class <- ifelse(!all(is.na(v.vals)), class(v.vals), out.class)
     } else {
       if (any(sapply(list, is.list))){
--- a/R/utils-list.R
+++ b/R/utils-list.R
@@ -30,23 +30,35 @@ remove_field <- function(list, param){
 
 strip_pts <- function(list, param){
   out <- c()
-  # out.class <- 'numeric'
+  out.class <- 'numeric'
   for (v in param){
     if (v %in% names(list) && !inherits(list[[v]], c('function','formula'))) {
       v.vals <- list[[v]]
       out <- append_keepTZ(out, v.vals, tz=attr(v.vals, 'tzone'))
-      out.class <- ifelse(!all(is.na(v.vals)), class(v.vals), out.class)
+      all.na <- all(is.na(v.vals))
+      out.class <- ifelse(!all.na, class(v.vals), out.class)
+      if(all.na){ 
+        class(out) <- out.class 
+      }
     } else {
       if (any(sapply(list, is.list))){
         u.list <- unname_c(list[sapply(list, is.list)])
         if(v %in% names(u.list)) {
           v.vals <- u.list[[v]]
           out <- append_keepTZ(out, v.vals, tz=attr(v.vals, 'tzone'))
-          out.class <- ifelse(!all(is.na(v.vals)), class(v.vals), out.class)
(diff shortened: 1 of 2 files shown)


R/utils-list.R
```diff
@@ -34,28 +34,40 @@ strip_pts <- function(list, param){
...
-          out.class <- ifelse(!all(is.na(v.vals)), class(v.vals), out.class)
+          out <- append_keepTZ(out, v.vals, tz=attr(v.vals, 'tzone'))
+          all.na <- all(is.na(v.vals))
+          out.class <- ifelse(!all.na, class(v.vals), out.class)
+          if(all.na){ 
+            class(out) <- out.class 
+          }
         } else {
-          out <- append(out, NA)
+          out <- append_keepTZ(out, NA, tz=attr(out, "tzone"))
```
jordansread, line 63: is it not necessary to set the class here too? Just wondering what cases we lose it for, and if this is a safe one

ldecicco-USGS: In my experience, POSIX's and Dates tend to keep their class, but they get attributes stripped off for various reasons. So, my guess is that this is fine.

lindsayplatt: That was only when all of "v.vals" was NA - we had to be explicit about what class it should have. Otherwise, it will have the same class as v.vals.

lindsayplatt:
@jread-usgs @ldecicco-USGS merge-ready?

R/utils-list.R
```diff
@@ -34,28 +34,40 @@ strip_pts <- function(list, param){
...
+        class(out) <- out.class 
+      }
     } else {
       if (any(sapply(list, is.list))){
         u.list <- unname_c(list[sapply(list, is.list)])
         if(v %in% names(u.list)) {
           v.vals <- u.list[[v]]
-          out <- append(out, v.vals)
-          out.class <- ifelse(!all(is.na(v.vals)), class(v.vals), out.class)
+          out <- append_keepTZ(out, v.vals, tz=attr(v.vals, 'tzone'))
```
ldecicco-USGS, line 48: Would there ever be a case where we were going to use some other tz? If not, should it just be moved into the function?:

```
append_keepTZ <- function(base.vals, append.vals){
 tz <- attr(append.vals, "tzone")   
 vals <- append(base.vals, append.vals)
   attr(vals,"tzone") <- tz
   return(vals)
 }
```

lindsayplatt: Some keep the tz to be from base.vals, and some from append.vals, so I kept it outside. Most need to look at the tzone of the values you are appending because you could be appending to an empty object. However, sometimes you are appending an NA (see L66), and then you need to keep the tzone from the existing object.

ldecicco-USGS:
So, I merged this, then remembered...it's always good to see that the gsplotIntro.html file has at least a minor update (because we show the date). Then we know that all the vignette and (ideally) readme plots look the same.

lindsayplatt:
Except that we don't commit the html file, so how could we tell if there was a difference?

ldecicco-USGS:
Why not? I do...

ldecicco-USGS:
Oh, I see @jread-usgs added it to the git ignore...I have no idea why. I consider it a part of the tests. I'm taking it off the git ignore

ldecicco-USGS:
No, actually he's right. It's ignored from the vignette folder, but not the "proper" location in inst/doc. So, before each PR...you should be doing:

```
devtools::build_vignettes()
```

and knitting the Readme.Rmd.

lindsayplatt:
just ran it - it did not actually change anything. I think it's because none of our examples have timezones, so they were showing up just fine.

ldecicco-USGS:
I didn't expect it to change anything...just mentioning that it's probably a good idea to always include it in your (and all of our) pull requests. At least the date up top changed though, right?

https://rawgit.com/ldecicco-USGS/gsplot/master/inst/doc/gsplotIntro.html

right under the author list, there's the date.

lindsayplatt:
oh yes - that changed. Worth making a new PR for though?

jordansread:
Agree @ldecicco-USGS, since we can't do visual tests, we should bug each other in PR reviews to include the knits

ldecicco-USGS:
Not worth another PR (I've actually got it in mine now....)...just mentioning for the future.

PR #416, opened 2016-11-08
Re-trying config branch

ldecicco-USGS:


ldecicco-USGS:
@lindsaycarr could you give this a good vetting? Maybe run your AQCU stuff? In the meantime, I'll try to figure out the best way to add some tests. Here's what I'm running locally:

superFat.yaml:

```
lwd: 3
lines: 
  lwd: 3
abline:
  col: "grey"
  lty: 1
  lwd: 2

```

```
df <- data.frame(x = 1:10, y=1:10, z = seq(2,20,2))
loadConfig("D:/LADData/RCode/gsplot/superFat.yaml")
gsp <- gsplot() %>%
  lines(df$x, df$y, col="red", legend.name = "points") %>%
  lines(df$x, df$z, legend.name = "lines") %>%
  legend()
gsp
```

```
loadConfig()
gsp <- gsplot() %>%
  lines(df$x, df$y, col="red", legend.name = "points") %>%
  lines(df$x, df$z, legend.name = "lines") %>%
  legend()
gsp
```

```
gspFAT <- gsplot(config.file = "D:/LADData/RCode/gsplot/superFat.yaml") %>%
  lines(df$x, df$y, col="red", legend.name = "points") %>%
  lines(df$x, df$z, legend.name = "lines") %>%
  abline(v=4) %>%
  legend()
gspFAT
```

```
gsp <- gsplot() %>%
  lines(df$x, df$y, col="red", legend.name = "points") %>%
  lines(df$x, df$z, legend.name = "lines") %>%
  legend()
gsp

```

Then, from our last pr:

```
g1 <- gsplot() %>% 
  lines(1:9, seq(1,3, by=0.25), 
  # formatting y axes
  axis(side=2, at=1:10) 

layout(matrix(1:2, 1, 2))
g1
g1
```

and

```
par(mar=c(5,4,10,2))

g1 <- gsplot() %>% 
  points(1:10, 1:10, pch=20, legend.name="first points") %>% 
  lines(4:1, 4:1, legend.name="first line") %>% 
  points(c(3,7,4), c(9,3,6), pch=20, col="black", legend.name="second points") %>% 
  legend(location="above")
g1
```

```

```

ldecicco-USGS:
(I don't have time to paste in the images, but I'll do that tomorrow too!....ideally you can imagine the first one has fat lines, the second is normal, the third fat, the fourth normal, then the last 2 are your examples)

lindsayplatt:
If you try that last one with the fat lines config, it loses the default color ('red') and seems to change the point size:

``` r
par(mar=c(5,4,10,2))
g1 <- gsplot() %>% 
    points(1:10, 1:10, pch=20, legend.name="first points") %>% 
    lines(4:1, 4:1, legend.name="first line") %>% 
    points(c(3,7,4), c(9,3,6), pch=20, col="black", legend.name="second points") %>% 
    legend(location="above")
g1
```

image

``` r
par(mar=c(5,4,10,2))
g1 <- gsplot(config.file = "D:/LRCData/R/gsplot/superFat.yaml") %>% 
    points(1:10, 1:10, pch=20, legend.name="first points") %>% 
    lines(4:1, 4:1, legend.name="first line") %>% 
    points(c(3,7,4), c(9,3,6), pch=20, col="black", legend.name="second points") %>% 
    legend(location="above")
g1
```

image

ldecicco-USGS:
I think that's right. The second example should lose the red color since it's not looking at the default config (which is where we are saying use red for points). 

The fatter points are consistent with base R:

```
plot(1:10, 1:10, type="p", pch=20, lwd=3)
```

rplot01

vs:

```
plot(1:10, 1:10, type="p", pch=20)
```

rplot02

lindsayplatt:
Ahh ok - I guess I was thinking of this as someone appending to or modifying the default config.

ldecicco-USGS:
oy vey, I suppose that will be something someone requests next. Let's start easy and add a "append" type argument later.

lindsayplatt:
When this ran with repgen, it passes all the tests but there is not legend appearing even though it is in the object. Might be a margin issue?

lindsayplatt:
[testing_legendissue.txt](https://github.com/USGS-R/gsplot/files/593238/testing_legendissue.txt)

It just seems to be when it's in an Rmarkdown that the legend can't showup. It also appears on my master branch, so this must not be associated with your changes...hunting to figure out when it did get changed.

ldecicco-USGS:
Hmm...seems like if you include the `fig.height=10, fig.width=13` in the chunk options, then it loses the legends. If you just specify the width, and let it figure out the height, I think it works.

lindsayplatt:
``` r
library(gsplot)
par(mar=c(5,4,10,2))

g1 <- gsplot() %>% 
  points(1:10, 1:10, pch=20, legend.name="first points") %>% 
  lines(4:1, 4:1, legend.name="first line") %>% 
  points(c(3,7,4), c(9,3,6), pch=20, col="black", legend.name="second points") %>% 
  lines(1:2, c(1,4), legend.name="second line") %>% 
  legend(location="above")
g1
```

Resetting to #400 does not ignore `mar`:
image

Resetting to #410 does ignore `mar`: 
image

@jread-usgs any ideas what it could be from? It uses par the second time to call `g1` if you don't sweep the plotting region first.

lindsayplatt:
Hmm but why would it have worked with specifying both before? We need to specify both fig.height and fig.width for repgen.

ldecicco-USGS:
Just so I understand what that last comment is talking about...This current pull request does not ignore the `mar` command in normal R scripting. The issue is that the behavior in an Rmd file has changed? But if you go back to #400, the Rmd knitting looks right?

lindsayplatt:
margin/par issue linked to #410. Investigating separately.

Issue #245, opened 2015-09-16
lost index for points and lines


jordansread:
``` R
gs <- gsplot()  %>% 
    points(1:20, col="blue", legend.name="Points 1")  %>% 
    points(20:1, col="green", legend.name="Points 2") %>% 
    legend(location="below")
gs
```

now fails

ldecicco-USGS:
Is this what you are talking about in #224?

lindsayplatt:
Do we need it to have the default labels - e.g. x-axis says "Index", then y-axis says "1:20"? My vote is no.

jordansread:
I think long ago we opted to not have gsplot do that. My vote is no as well.

PR #420, opened 2016-12-30
indexing allowed for points and lines

lindsayplatt:
Does not label x axis as "Index" and y with range. Works for logged axes and keeps user-specified axis labels. Also works with dates and other classes. It also works for `lines`.

See #245 

Originally, I had implemented this using `xy.coords` since that is what `points.default` and `lines.default` call, but if coerces data into numeric. 

```r
gs <- gsplot()  %>%
    points(1:10, xlab="my points", log="y")
gs
```

image

### DIFF
--- a/R/set_args.R
+++ b/R/set_args.R
@@ -15,7 +15,7 @@ set_args <- function(fun.name, ..., custom.config = FALSE, package='graphics'){
   user_args <- function_args(name=fun.name, package=package, ...)
   
   if(fun.name %in% c('points', 'lines')){
-    xy_args <- xy.coords(x = user_args$x, y = user_args$y)
+    xy_args <- xy.coords(x = user_args[['x']], y = user_args[['y']])
     user_args <- append_replace(user_args, xy_args)
   }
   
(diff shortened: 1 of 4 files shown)


R/set_args.R
```diff
@@ -14,6 +14,14 @@ set_args <- function(fun.name, ..., custom.config = FALSE, package='graphics'){
   config_args <- config(fun.name, custom.config = custom.config)
   user_args <- function_args(name=fun.name, package=package, ...)
   
+  if(fun.name %in% c('points', 'lines')){
```
jiwalker-usgs: I think this should maybe be in function_args rather than here.  I worry that adding special cases in set_args will complicate this function beyond where it is now.

jiwalker-usgs (approved):
I think this is fine, I'm just wondering if we will have an issue with functions like this going forward.

jordansread:
I agree w/ what @jiwalker-usgs said. this looks fine to me too, but we should (as in the past) just be able to pass the args on to points w/ y=NULL and have that function handle it appropriately.

jordansread (approved):

lindsayplatt:
@jread-usgs have points.gsplot handle it appropriately or graphics::points? The issue seems to be that we never pass it to points.default which calls plot.xy and handles this issue of y=NULL.

jordansread:
I meant `graphics::points`, but I guess maybe this never worked correctly then?

lindsayplatt:
I guess I'm struggling with the `override` function and how `UseMethod` works. It should get to points.default somehow, which handles the xy indexing https://svn.r-project.org/R/trunk/src/library/graphics/R/points.R

Issue #304, opened 2016-01-21
add creation/config metadata to gsplot object


jordansread:
i.e., this object was created w/ `gsplot` version x and extension y...

lindsayplatt:
just add gsplot version for this issue, creation date

PR #421, opened 2016-12-30
Add metadata

lindsayplatt:
Now each gsplot object will have a top-level list called `metadata`. This PR adds 2 fields: `created` which is the date the object is/was created and `gsplot.version` the gsplot version used to create the object. Both can be manually input as arguments to `gsplot()`. 

This PR also adds to access functions to get those metadata values - `whatDate` and `whatVersion`.

See #304


R/gsplot-class.R
```diff
@@ -19,8 +23,12 @@ gsplot <- function(x = NULL, ...) UseMethod("gsplot")
 
 #' @rdname gsplot
 #' @export
-gsplot.default <- function(...,config.file=NA, theme=NA, frame.plot=TRUE) {
-  object <- gsplot(list(global=list('config'=list(frame.plot=frame.plot, 
+gsplot.default <- function(..., created=Sys.Date(), 
+                           gsplot.version=packageDescription("gsplot", fields = "Version"),
```
jiwalker-usgs: might use `packageVersion("gsplot")` here instead

jiwalker-usgs (approved):
I like this, I'm wondering if there is other metadata we should consider.  Also, are there other accessor methods that could follow the what pattern here?

R/gsplot-class.R
jordansread: better yet is 
```r
packageVersion(getPackageName())
```
so we don't need to modify it if we change the name of the package.

jordansread (approved):

ldecicco-USGS:
What is this magic "Approved" that you guys are doing?

ldecicco-USGS (approved):
Ahhh! Cool.

R/gsplot-class.R
lindsayplatt: I tried `packageVersion("gsplot")` at first, but it returns a list instead of a character string. `packageDescription("gsplot", fields = "Version")` returns a single string. To get the output of packageVersion into a single string seemed more complex than using packageDescription. 

I will make the package reference generic per @jread-usgs's suggestion.

lindsayplatt:
@jiwalker-usgs I think there were other metadata that people suggested, but we decided that #304 would just be the gsplot version and date.

lindsayplatt:
@jread-usgs the appveyor build failed...is that OK?

PR #425, opened 2017-02-01
Theme

ldecicco-USGS:


ldecicco-USGS:
Hmm...after building my readme and vignette, I see there's a par thing still going on.

jiwalker-usgs:
I'd prefer the code that builds the theme objects in source rather than just in the sysdata.rda (more transparent and maintainable).

ldecicco-USGS:
Hmm...so then would you not want the thing that is assigned to the object? I'm confused. 

 I want the user to also be able to make a theme easily in their workspace..so....I'll have to think how that could all be done the way you are asking. As it stands, you can use a theme that's packed up in the package, or make one yourself:


```
theme.packers <- gsplot() %>%
  bgCol(col="#203731") %>%
  grid(col = "gold", lty = 1, lwd=3) %>%
  par(tcl=-0.2, cex=0.75) %>%
  axis(side = c(1,2), lwd = 0, lwd.ticks = 1)

save(theme.packers, "packers.RData") # To use it again someday

gsplot(theme=theme.packers) %>%
  points(1:4, 1:4, col="brown", pch=20, cex=3)

```
image

jiwalker-usgs:
Yeah, that way would work, I'm just thinking that the themes that are built into the package should be defined in source.

ldecicco-USGS:
I'm still confused, but I'm sure you can explain it to me in person.

ldecicco-USGS:
(i like the packer theme....that might have to go in there)

R/gsplot-class.R
```diff
@@ -27,11 +35,27 @@ gsplot.default <- function(..., created=Sys.Date(),
...
+                                      theme[["global"]][names(theme[["global"]])[names(theme[["global"]]) != "config"]])),
+                            theme[names(theme)[!(names(theme) %in% c("metadata","global"))]]
+                          )
+                     )
+  } else {
+    object <- gsplot(list(metadata=list(created=created,
+                                        gsplot.version=gsplot.version),
+                          global=list('config'=list(frame.plot=frame.plot, 
+                                                    config.file=!is.na(config.file)))))    
+  }
```
lindsayplatt: rather than set global and metadata in two places, create theme object OR gsplot object THEN add global/metadata

R/gsplot.R
```diff
@@ -70,3 +70,17 @@ NULL
...
+#'}
+#'
+#'@aliases theme.hadley
+#'@name Themes
+#'@docType data
+#'@export theme.hadley
+#'@keywords datasets
+#'@examples
+#' ggplotTheme <- theme.hadley
+NULL
```
lindsayplatt: put theme object here rather than in sysdata.rda to expose how it was created

R/gsplot-class.R
```diff
@@ -27,22 +40,46 @@ gsplot.default <- function(..., created=Sys.Date(),
...
+       "config.file" %in% names(theme[["global"]][["config"]])){ #if no config file specified by user
+      config.file <- theme$global$config$config.file
+    }
+    
+    object[["global"]][["config"]]["config.file"] <- config.file
+    
+  } else {
+    object <- list(global= list(config=list(frame.plot=frame.plot,
+                                            config.file = !is.na(config.file))))
+  }
```
lindsayplatt, line 63: `append_replace` could help w/ if-else's

lindsayplatt:
Partially fixing #425. Still working on testing config to theme override behaviors.

User config has priority over theme config. Only change theme config settings if user-config has them.

Issue #417, opened 2016-11-09
config function only sets args in "vanialla" graphics-like functions


ldecicco-USGS:
I think the config file (both via `loadConfig` and using a temporary config file are only changing `par`. So if it's a non-par argument, it's ignored (with a warning at least). For example....`label=FALSE`. `label` is an argument of `axis`, but not a `par` argument. 

```
loadConfig(system.file("extdata", "lineScatter.yaml", package = "gsplot"))
df <- data.frame(x = 1:10, y=1:10, z = seq(2,20,2))

gspDef <- gsplot() %>%
                  points(df$x, df$y, legend.name = "points") %>%
                  bgCol() %>%
                  legend()

gspDef
```

![rplot](https://cloud.githubusercontent.com/assets/1105215/20145661/94011b14-a666-11e6-8366-dad372728f63.png)

The lineScatter.yaml has the `bgCol` set to lightgrey, and the title of the legend should be "EXPLANATION".

My understanding was that we wanted this file (or gsplot options) to be the default arguments for any function (not just par). So, either it's a bug, or I understand the config file wrong. In which case, we should re-jigger the documentation and yaml files to only have par arguments.

ldecicco-USGS:
So, legend, error_bars, and callouts I think are all missing this functionality.

ldecicco-USGS:
To do:
- [x] error_bar
- [x] callouts
- [x] legend
- [x] bgCol (rename to background_color)

PR #424, opened 2017-02-01
Axis log wrong side --- FIX

lindsayplatt:
Fixing #414 

Test out:

```r
# side 4 logged, side 2 not logged
gs <- gsplot() %>% 
  points(1:3, c(1,10,100)) %>% 
  points(3:5, c(10,100,1000), col="blue", side=4, log='y') %>% 
  axis(side=4)
gs

# side 3 logged, side 1 not logged
gs <- gsplot() %>% 
  points(c(1,10,100), 1:3, side=1) %>% 
  points(c(10,100,1000), 3:5, col="blue", side=3, log='x') %>% 
  axis(side=3)
gs

# side 3 logged, side 1 not logged
gs <- gsplot() %>% 
    points(c(1,10,100), 1:3, side=1) %>% 
    points(c(10,100,1000), 3:5, col="blue", side=3) %>% 
    axis(side=3) %>% 
    view(side=3, log='x')
gs

# side 1 logged, side 3 not logged
gs <- gsplot() %>% 
    points(c(1,10,100), 1:3, side=1) %>% 
    points(c(10,100,1000), 3:5, col="blue", side=3) %>% 
    axis(side=3) %>% 
    view(side=1, log='x')
gs
```

PR #428, opened 2017-02-04
error_bar fix

ldecicco-USGS:
Is a start on #417

Still don't know what triggers a transparent background on my computer.


R/error_bars.R
```diff
@@ -38,29 +37,99 @@
 #' 
 #' gs <- gsplot() %>%
 #'    points(1:10, 1:10) %>%
-#'    error_bar(5, 5, y.high=1) 
+#'    error_bar(x=5, y=5, y.high=1, col="green")
```
jiwalker-usgs: are the x and y still required?

R/error_bars.R
```diff
@@ -38,29 +37,99 @@
...
+#' @param x location in x
+#' @param y location in y
+#' @param y.high offset up
+#' @param y.low offset down
+#' @param x.high offset right
+#' @param x.low offset left
+#' @param epsilon width of bar in relative units
+#' @param \dots additional arguments passed to \code{\link[graphics]{arrows}}
+#' @export
+#' @keywords internal
```
jiwalker-usgs: should this be internal?

R/error_bars.R
```diff
@@ -38,29 +37,99 @@
...
+#'    error_bar(x=5, y=5, y.high=1, col="green") 
 #' gs 
 error_bar <- function(object, ...) {
   override("gsplot", "error_bar", object, ...)
 }
 
 
-error_bar.gsplot <- function(object, x, y, y.high=0, y.low=0, x.high=0, x.low=0, 
-                             epsilon=0.1, ..., legend.name=NULL, side=c(1,2)){
+error_bar.gsplot <- function(object, ..., legend.name=NULL, side=c(1,2)){
```
jiwalker-usgs, line 47: I feel like the signature for this should be what it was, it is redundant in a way as it is mirrored in the default, but it will help with the argument handling.

jiwalker-usgs:
We are back to transparent plots, which is right?

ldecicco-USGS:
I'm trying to figure out the transparency thing. It might be the order on if I build the vignette first and then the readme.

R/error_bars.R
ldecicco-USGS: no

R/error_bars.R
ldecicco-USGS: not anymore, I'll bring it out

lindsayplatt:
If it's the order that is making the difference, then would that be a par resetting issue?

ldecicco-USGS:
That's my theory, I haven't had time to check yet.

R/error_bars.R
ldecicco-USGS: I tried hard to get those defaults to work. It's not as simple as listing the y.high's and others along with the `...`, I think because of that lazy eval? I'll give it another go....maybe I missed something the first time I tried. Currently, if I just do this:

```
fun.name='error_bar'
  
  args <- list(x=x, y=y, y.high=y.high, y.low=y.low, x.high=x.high, x.low=x.low, epsilon=epsilon, ...)
    
  object <- gather_function_info(object, fun.name, args, legend.name=legend.name, side=side)
  
```

(instead of what's in the PR right now), I get the error:
```
Error in is.finite(side.vals) : 
  default method not implemented for type 'list'
```
coming back from `gather_function_info`

ldecicco-USGS: Ha! was making my list wrong. @jiwalker-usgs have a look. I'll try to add some new tests to get that coveralls at least back up to even.

ldecicco-USGS: Oh shoot, no, this still doesn't work.

ldecicco-USGS:
Building the vignette and looking at it, something looks amiss with `layout`, I'll have a look later. That might have nothing to do with this particular PR.

R/error_bars.R
ldecicco-USGS: Now it works

ldecicco-USGS:
I've confirmed the `layout` issue happened before this PR (I'm pretty sure it was on #424. So, who merged that without checking the vignette? Oh yeah, me....lesson learned). So, I added the the offending graph to the README so we can see it and fix it easier in the future.

PR #431, opened 2017-02-08
Layout bug & more!

ldecicco-USGS:
* Adds limits to abline for v and h
* https fixes
* fixes some bugs

So now, what works (that didn't in last 2 PRs):
```
g1 <- gsplot(xpd=FALSE) %>% 
  lines(1:9, seq(1,3, by=0.25), 
  axis(side=2, at=1:10)
g1
layout(matrix(1:2, 1, 2))
g1
g1
```
image

```
#Works:
gs1 <- gsplot() %>% 
  points(1:3, c(1,10,100)) %>% 
  points(3:5, c(10,100,1000), col="blue", side=4, log='y') %>% 
  axis(side=4) 
gs1
```
image

```
#Works:
gs2 <- gsplot() %>% 
  points(c(1,10,100), 1:3, side=1) %>% 
  points(c(10,100,1000), 3:5, col="blue", side=3, log='x') %>% 
  axis(side=3) 
gs2
```
image

UPDATE!!!
```
#Works:
layout(matrix(1:2, 1, 2))
gs1
gs2
```
image

and also still doesn't work:
```
gs <- gsplot() %>% 
  points(c(1,10,100), 1:3, side=1) %>% 
  points(c(10,100,1000), 3:5, col="blue", side=3) %>% 
  axis(side=3) %>% 
  view(side=1, log='x')
gs
```
image

only side 1 should be logged, not 3

I think those are 2 current bugs, but this PR gets us a bunch fixed as-is. 

I'll try to get these bugs listed as issues.


R/print.R
```diff
@@ -62,12 +62,20 @@ print.gsplot <- function(x, ...){
     }
 
     print.view(views[[view.name]])
-    par(old.par)
+    
+    
+    if(is.na(as.logical(all.equal(c(1,1), par()$mfrow))) & is.na(as.logical(all.equal(c(1,1,1,1), par()$mfg)))){
```
ldecicco-USGS: This line is pretty much the meat of this PR. Also where the bug fix will need to happen to get the right log scales in a layout.

ldecicco-USGS:
Oh, and I added a function `clear_par` that will reset to the par that was on the users computer when they loaded the package. I think it's useful...not sure if we need something fancier, or if that's it.

lindsayplatt:
@ldecicco-USGS those log separate sides issues were fixed in #424. This seems to break that behavior?

Based on #424:
```r
gs <- gsplot() %>% 
  points(c(1,10,100), 1:3, side=1) %>% 
  points(c(10,100,1000), 3:5, col="blue", side=3) %>% 
  axis(side=3) %>% 
  view(side=1, log='x')
gs
``` 
returns side 1 logged, but not side 3 as I would expect
image

ldecicco-USGS:
Yeah, but that's the PR that broke all the `layout` graphs. This pr fixes much of layout and 3/4 log axis graphs.

ldecicco-USGS:
Currently struggling with:
```
view.usr <- par('usr')
```


unnamed-chunk-22-1.png shows the issues.

jiwalker-usgs:
The second plot in the diff seems to have issues with it (legend overlaps y axis).  Otherwise this looks pretty good.

ldecicco-USGS:
yeah, I think the legend placement is because of that `par('usr')`...still fiddling with it.

ldecicco-USGS:
A-OK! I think a bunch of stuff is fixed by doing a final `set_frame` call towards the end of `print` to make the default frame at the end view.1.2 (if defined), or the first one that the user defined (if 1.2 isn't defined).


I think it's good-to-go.

jiwalker-usgs (approved):

PR #434, opened 2017-02-10
Date axis

jiwalker-usgs:
Adds some utilities to handle special date axes.  Not sure it handles all the cases we want to deal with, but this is hopefully a good start.


ldecicco-USGS:
I just took it for a test ride, looks good! I'd say it's ready to merge. I'm not sure if this would be appropriate for this PR, or a new feature, but I would like to do something like:

```
x <- seq(as.Date("2013-01-22"), as.Date("2013-10-02"), "days")
y <- rnorm(length(x), 71, 19)
gs <- gsplot() %>%
  points(x, y) %>%
  date_axis(side=1, pos.lab="interval", tick.int="month", snap.to="month") %>%
  axis(side=c(3,4), labels = FALSE) %>%
  grid()
gs

```

And have the ticks line up on side 3 with side 1. What I see is:
image

jiwalker-usgs:
Yeah, that definitely should work, I'll make an issue as it is probably a bit involved to be able to inherit lazy eval'd limits.

PR #435, opened 2017-02-13
Grid fix

ldecicco-USGS:
Fix with grid, and a bunch more examples in the readme. This (I think) clears up some config/theme questions...very basically....My next plan is to revamp our examples in the vignette to make them a bit more user-friendly.


ldecicco-USGS:
FYI...the png that has the difference in line width on error_bars is correct. I wanted to use the "embed" function to clarify what we had to offer, and slightly simplify the graph.

R/gsplot-class.R
```diff
@@ -52,22 +52,24 @@ gsplot.default <- function(..., created=Sys.Date(),
     
     if(is.na(config.file) && 
        "config.file" %in% names(theme[["global"]][["config"]])){ #if no config file specified by user
-      config.file <- theme$global$config$config.file
+      config.file <- theme$global$config$config.path
     }
     
-    object[["global"]][["config"]]["config.file"] <- config.file
+    object[["global"]][["config"]][["config.file"]] <- !is.na(config.file)
+    object[["global"]][["config"]][["config.path"]] <- config.file
```
jiwalker-usgs, line 59: What is the difference between config.file and config.path?

jiwalker-usgs (approved):

R/gsplot-class.R
ldecicco-USGS: One is a logic that says either use the custom config or not. The other is the path to the custom config file. We didn't need the path before themes because the custom config defaults were stored in a temporary gsconfig environment. I was thinking about this yesterday, and I think I'm going to change it to save custom config parameters directly in the gsplot object. Then, if someone makes a theme with a config file, that theme can be transported around no problem. So...stay tuned, I think I can get that in this PR sometime today.

PR #439, opened 2017-02-14
Version update

ldecicco-USGS:


ldecicco-USGS:
Just going to update the version since there were 2 fairly big PR's since our last version update

PR #441, opened 2017-02-15
Theme test fix

ldecicco-USGS:


lindsayplatt:
@ldecicco-USGS what am I testing here? Is there something specific I need to look for?

ldecicco-USGS:
I got trigger happy on the PR #435  (it was approved to merge, but I pushed an extra commit that I forgot I pushed...). Anyway, that new commit broke a test, now I fixed that, and that's what this is. I'd say, if you want to just pull down this branch and see if this works on your computer, we should be good:

```
gs_packers <- gsplot(theme = theme.packers) %>%
  points(1:10, 1:10)
gs_packers

```

lindsayplatt:
Is the documentation for the `theme` arg still accurate? 

`path to the file that will only be used for setting the gsplot theme in this one gsplot object.`

It seems like there are many more options for theme than what it says: use a built-in theme by using the object name, create your own theme object and put that here, or put a path to a .Rdata file that specifies your theme. It would be nice to update that. But maybe that can go with a diff pr

lindsayplatt (approved):

ldecicco-USGS:
You're right, documentation should be improved.

ldecicco-USGS:
OK, updated the documentation. If it looks OK, go ahead and merge. Across the board, I think we need to work on our documentation, but that will be for milestone 0.9.0

lindsayplatt:
Travis seems to be stuck again. I'm going to go ahead and merge since it passed the AppVeyor build

PR #445, opened 2017-02-16
Axis stuff

ldecicco-USGS:
This PR currently fixes gsplot in that it will actually work (what is currently on the USGS-R master doesn't work at all). BUT! I commented out some tests, specific to `append` on the axis. 

So...we could merge this one right away if we want the master on USGS-R to be functional asap. Or, I might be able to uncomment these failing tests in the next day or so to get append working again. Anyway, I'd vote merging this for now on the off-chance someone is installing via install_github.


ldecicco-USGS:
I'm a bit confused right now. I've got other files that are suppose to be included on this, but they aren't showing up on github. (and when I push, it says everything is up-to-date). So...there might be a glitch in the github matrix currently. Hold on...

ldecicco-USGS:
Aaaaaand there they are! Weird...

Issue #437, opened 2017-02-14
minor ticks aren't inheriting pars


ldecicco-USGS:
```
gsplot(frame.plot=FALSE) %>%
   background_color(col="grey90") %>%
   grid(col = "white", lty = 1) %>%
   par(tcl=-0.2, cex=0.75, xaxs= "r", yaxs =  "r") %>%
   axis(side = c(1,2), lwd = 0, lwd.ticks = 1, n.minor=1)
```
Should have the minor tick marks going outside, shows up inside:

![image](https://cloud.githubusercontent.com/assets/1105215/22940508/b4bee3a8-f2a7-11e6-8194-a9c2f442e57f.png)

ldecicco-USGS:
Also:
```
gs <- gsplot() %>%
  points(1:100, 1:100, log="xy") %>%
  axis(side=c(3,4), labels=FALSE, n.minor=4) %>%
  axis(side=c(1,2), n.minor=4)
gs

```
side's 3 and 4 ticks aren't logged (and should be).

lindsayplatt:
:point_up: @ldecicco-USGS I'm assuming this is example is something to do with #436 ?

ldecicco-USGS:
They might be related, but that issue is specific to the date_axis.

ldecicco-USGS:
Actually, that second example I posted has nothing to do with minor ticks. This:

```
gsplot() %>%
   points(1:100, 1:100, log="xy") %>%
   axis(side=c(3,4)) 
```

Shows a linear axis on 3 and 4 (and log on 1, 2). I would expect 3 and 4 to be logged if no other call to 3 & 4 are defined.

lindsayplatt:
Yeah, that second example seems like a second issue.

ldecicco-USGS:
Yup, making a new issue.

lindsayplatt:
I'm seeing logged ticks now, but still seeing minor ticks on the inside instead of outside:

```r
gsplot(frame.plot=FALSE) %>%
   background_color(col="grey90") %>%
   grid(col = "white", lty = 1) %>%
   par(tcl=-0.2, cex=0.75, xaxs= "r", yaxs =  "r") %>%
   axis(side = c(1,2), lwd = 0, lwd.ticks = 1, n.minor=1)
```
![image](https://cloud.githubusercontent.com/assets/13220910/24115142/71e10f78-0d70-11e7-8241-b13e6978148f.png)


```r
gs <- gsplot() %>%
[...]

PR #460, opened 2017-03-22
tcl.minor default

jordansread:
for #437

I ended up doing this in a way that isn't testable except for the result of the render...

This is because I don't want the fraction of `par('tcl')` to be calculated until it is relevant - i.e., when the side is being rendered, so it is stuck into the `draw_axis` w/ the arg = `NA` signifying to use this default. It is buried from the user, but the alternative solution that covers this and was discussed in #437 with the hierarchy is too complex for the value of this functionality. Additional thoughts welcome.


lindsayplatt:
Since it's not testable except for rendering, did you build any of the vignettes? I imagine we would see a change because the `tcl.minor` default was 0.15 and should now be -0.25 if `par('tcl')` remains unchanged.

I think having NA carry through to the final gsplot object makes sense to me. People can just look in the documentation to figure out what the actual value should be since it could change if pars change.

lindsayplatt (changes requested):
Change the gsplot config to have a default `tcl=0.5` instead of the default par which is `tcl=-0.5`, so that ticks remain on the inside.
Build the vignettes and README first.

jordansread:
@lindsaycarr the `tcl` is in the default config already as 0.3. I will keep that as-is. Working on merge conflicts for the vignettes now and that is coming w/ the next commit.

jordansread:
changes in the vignettes all seem to be due to the retina screen render. Note we don't have anything in the readme that calls out the tcl.minor, so we don't see changes there.

lindsayplatt:
Doesn't it appear that the new fig changes are overplotting axes? Or is that just a visual trick?

jordansread:
Hard to say if overplotting or retina. I didn't change anything related to append vs replace for axis, so not sure why this would overplot if it didn't before this commit.

jordansread:
I added a change to the readme to force it to use the `tcl.minor` default @lindsaycarr

lindsayplatt (approved):

PR #461, opened 2017-03-22
Use travis R

ldecicco-USGS:


ldecicco-USGS:
Note: this very first one could take *awhile*, but with `cache: packages` the builds get speedier.

ldecicco-USGS:
So....why did changing the travis file make appveyor not kick off? Maybe it's super smart and realizes thing that it cares about changed (because we Rbuildignore the travis file)...or maybe appveyor's just being flaky again.

Issue #453, opened 2017-03-21
If side 1 or 2 isn't defined with data, can't modify axis.


ldecicco-USGS:
```
gsplot() %>%
     points(1:100, 1:100, log="xy", side=c(3,4)) %>%
     axis(side=c(1), log="") 
```

![image](https://cloud.githubusercontent.com/assets/1105215/24163357/53527bbe-0e38-11e7-965c-6f74545bb194.png)

Trying to figure out a way to not log side 1.

jordansread:
per #456 , `log` is being ignored here because it isn't a formal for any of the `axis` functions that gsplot looks to for arg names. My fix for #456 (WIP) causes your call to warn with
```
Warning message:
In axis(side = side, at = at, labels = labels, ...) :
  "log" is not a graphical parameter
```
which I think is right. 

Do you want 
```r
gsplot() %>%
    points(1:100, 1:100, log="xy", side=c(3,4)) %>%
    view(log='', side=c(1,2))
    axis(side=c(1)) 
```
That does the same thing but is the way you could modify the log status of 1 and 2 if there were data there. 

weirder yet is that this seems to ignore logging by tying the axes together too strongly:

```r
gsplot() %>%
    points(1:100, 1:100, log="xy", side=c(3,4)) %>%
    points(1:10,1:10, log="", side=c(1,2)) %>% 
    axis(side=c(1)) 
```
![image](https://cloud.githubusercontent.com/assets/2349007/24201792/1c3f7472-0edf-11e7-8829-cc3b0b36fbd3.png)

Do we have tests hitting this code?
https://github.com/USGS-R/gsplot/blob/455298e9de2b6dcfda7a9b9232d4c2b9f8727ce7/R/modify_side.R#L31

ldecicco-USGS:
I think we really need to emphasize the `view` function in the readme and vignette, since it's so critical and completely unlike base R. I would never think to use the view function, but makes sense once I see it.

lindsayplatt:
I agree, going to add an issue to add view as documentation in the README and vignettes. See #462

lindsayplatt:
What's the plan for finishing this one up @ldecicco-USGS @jread-usgs ? The axis issues for axis style args not getting recognized were fixed between #459 and #460. Do we just need to use view here instead and make sure to document that? Or are there code changes that need to take place?

jordansread:
no, I think this is a bug. 

```r
gs <- gsplot() %>%
[...]

Issue #414, opened 2016-11-07
log axis on wrong side


lindsayplatt:
gsplot object is correct, but plotting the wrong thing

- Might have something to do with copy if not explicit (JW)
- Maybe because print is weird???

```
# this should only have side 4 logged, the helper functions say the right thing, but the plot shows both logged
gs <- gsplot() %>% 
    points(1:2,c(2,25)) %>% 
    points(c(1,2), c(1, 1000000000), side=4, log='y') 
gs

logged(gs, side=2)
[1] FALSE
logged(gs, side=4)
[1] TRUE
```

![image](https://cloud.githubusercontent.com/assets/13220910/20079274/890b3810-a509-11e6-8d63-5005561c1daa.png)

lindsayplatt:
In print, in the side for loop we don't want to set `old.par <- par(x[[side.name]]$par)` . This should be done in two steps, or par() should return the old par before it was modified.

lindsayplatt:
I would like to be able to log each side separately, but that pops up with warnings currently. This should log the right axis, but not the left. I think it might be related to this issue.

``` r
gs <- gsplot() %>% 
    axis(side=4) %>% 
    view(side=2, ylim=c(1,100)) %>% 
    view(side=4, ylim=c(1,100), log='y') 

Warning messages:
1: In doTryCatch(return(expr), name, parentenv, handler) :
  nonfinite axis limits [GScale(-inf,0,2, .); log=1]
2: In doTryCatch(return(expr), name, parentenv, handler) :
  nonfinite axis limits [GScale(-inf,0,2, .); log=1]

gs

Warning message:
In plot.window(xlim = lim(object, side), ylim = lim(object, side),  :
  nonfinite axis limits [GScale(-inf,0,2, .); log=1]
```

jordansread:
does it work if you have data on those sides?

lindsayplatt:
Oh I guess it eliminates the warnings, but it logs both axis instead of just one. It's also not assigning the correct side as logged.

``` r
gs <- gsplot() %>% 
    points(1:2, c(1,100), side=2) %>% 
    points(3:4, c(10,1000), side=4) %>% 
    axis(side=4) %>% 
    view(side=2, ylim=c(1,100)) %>% 
    view(side=4, ylim=c(1,100), log='y') 
gs

> logged(gs, side=1)
[1] FALSE
> logged(gs, side=2)
[1] TRUE
> logged(gs, side=4)
[1] FALSE
```

[...]

PR #468, opened 2017-03-27
Axis stuff

ldecicco-USGS:


tests/testthat/tests-date_axis.R
```diff
@@ -21,22 +21,27 @@ test_that("axis ticks in the right location", {
...
+    points(seq(as.Date("2013-01-01"), as.Date("2013-12-31"), "days"), 1:365) %>%
     date_axis(side=1, lab.pos = "interval", tick.int="month", "snap.to"="month")
   
-  # labels <- lazyeval::lazy_eval(gs$side.1[[2]]$at, data=list(object=gs))
-  # expect_true(all.equal(labels[1], as.Date("2013-01-15"), tolerance=0.01))
-  # expect_true(all.equal(labels[7], as.Date("2013-07-15"), tolerance=0.01))
-  # expect_true(all.equal(labels[12], as.Date("2013-12-15"), tolerance=0.01))
+  second.axis <- gs$side.1
+  which.axis <- which(names(second.axis)== 'axis')
+  labels <- lazyeval::lazy_eval(gs$side.1[[which.axis[2]]]$at, data=list(object=gs))
```
ldecicco-USGS, line 29: @jiwalker-usgs I had to add this calculation (which.axis) because the order got jiggered around. It doesn't seem like that would affect your code...but please double check that things are working.

jiwalker-usgs: Yeah, this should be fine, I just made a brittle test.

ldecicco-USGS: Phew :sweat:

lindsayplatt:
Fixing #453. The following examples work with this PR:

```r
# this should give a warning - you can't use "log" inside an axis call
gsplot() %>%
     points(1:100, 1:100, log="xy", side=c(3,4)) %>%
     axis(side=c(1), log="") 
Warning message:
In axis(side = side, at = at, labels = labels, ...) :
  "log" is not a graphical parameter

# this is how you would have side 3 and 4 logged, but not 1
gsplot() %>%
    points(1:100, 1:100, log="xy", side=c(3,4)) %>%
    view(side=c(1,2), log='') %>% 
    axis(side=1) 

# this should log sides 3 & 4, but not 1 & 2
gsplot() %>%
    points(1:100, 1:100, log="xy", side=c(3,4)) %>%
    points(1:10,1:10, log="", side=c(1,2))
```

This is somewhat a continuation of #414.
```r
# this has sides 1,2 not logged and side 4 logged (no side 3)
gs <- gsplot() %>% 
    points(1:2,c(2,25)) %>% 
    points(c(1,2), c(1, 1000000000), side=4, log='y') 
gs
```

lindsayplatt (approved):
I added some examples of what was causing issues and how they should look as a comment for future reference. This looks good to me!

PR #473, opened 2017-04-17
increase version for AQCU

lindsayplatt:
I always get too excited and forget to do this along with the bug fix....


lindsayplatt:
Self-merging since this is just a minor version update.

